PR #22, opened 2015-05-29
Correct URL for Pull Requests

richrace:
Pull Request notifications had the URL: `https://www.github.com/atom/find-and-replace/pulls/399`

Which would link to an author. This PR corrects it to be: `https://www.github.com/atom/find-and-replace/pull/399`

I only had one notification to test it with though.


ekonstantinidis:
Hey @richrace ! Yes you're right! Was about to fix this one exactly the same way! :rocket:

PR #34, opened 2015-05-30
Include Homebrew-Cask installation instructions.

eivind88:
Ammend readme to include instructions for installation via Hombrew-Cask.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,13 @@ GitHub Notifications on your menu bar.
 ### Installation
 You can download Gitify from the [releases](https://github.com/ekonstantinidis/gitify/releases) page. Currently only supports OS X.
 
+Gitify can also be installed using [homebrew cask](https://github.com/phinze/homebrew-cask).
+
+```
+brew cask install gitify
+```
+
+
 ### Prerequisites
 
  - Electron [+](http://electron.atom.io/)


ekonstantinidis:
Just had a look at `homebrew-cask` and how it works. So you install the app from the command line but you also have to use the command line to launch it (`open ~/Applications/"Gitify.app"`). To be honest I'm not convinced by the way it works and that we have to open a PR for every release. Was thinking that we can keep it the homebrew-cask release "unofficial" - no readme info? What do you think?

eivind88:
It's not made as clear as I think it should be, but you can ```export HOMEBREW_CASK_OPTS="--appdir=/Applications"``` in ```.bash_profile``` to symlink the installed apps into ```/Applications``` instead of```~/Applications```.
This will make the the apps appear in the same Application-folder as the one that's probably in your dock and your Finder sidebar, and in Spotlight search results.

I agree that the workflow of "PR-for-every-release" seems cumbersome.
There are [ways](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#adding-a-cask) around this, but a permanent link for the latest release binary would be needed. Github doesn't support this as far as i know.

I think "unofficial" homebrew-cask releases seems like a good solution.

ekonstantinidis:
I had a look at the examples and what would be ideal is what Firefox did ([example](https://github.com/caskroom/homebrew-cask/blob/master/CONTRIBUTING.md#examples)) - adding a cask that will always grab the latest release but this is not possible with GitHub. So I'm closing this one and we're going with "unofficial" homebrew-cask releases! Thanks for letting me know!

PR #41, opened 2015-05-31
Works on Ubuntu now.

caske33:
This should be tested on a OS X too!

I changed the windowsize a bit because there were scrollbars when using the current windowsize.

### DIFF
--- a/main.js
+++ b/main.js
@@ -25,8 +25,8 @@ app.on('ready', function(){
 
   function initWindow () {
     var defaults = {
-      width: 400,
-      height: 350,
+      width: 420,
+      height: 370,
       show: false,
       frame: false,
       resizable: false,
@@ -41,9 +41,14 @@ app.on('ready', function(){
   }
 
   function showWindow (bounds) {
+    var ElectronScreen = require('screen');
+    var workAreaSize = ElectronScreen.getPrimaryDisplay().workAreaSize;
+    var windowSize = appIcon.window.getBounds();
+    var defaultX = workAreaSize.width-windowSize.width;
+    var defaultY = 0;
     var options = {
-      x: bounds.x - 200 + (bounds.width / 2),
-      y: bounds.y,
+      x: defaultX || bounds.x - 200 + (bounds.width / 2),
+      y: defaultY || bounds.y,
       index: path.join('./', 'index.html')
     };
 
@@ -93,6 +98,6 @@ app.on('ready', function(){
     app.quit();
   });
 
-  app.dock.hide();
+  //app.dock.hide();
   appIcon.setToolTip('GitHub Notifications on your menu bar.');
 });
--- a/main.js
+++ b/main.js
@@ -98,6 +98,8 @@ app.on('ready', function(){
     app.quit();
   });
 
-  //app.dock.hide();
+  if(process.platform === 'darwin'){
+    app.dock.hide();
+  }
   appIcon.setToolTip('GitHub Notifications on your menu bar.');
 });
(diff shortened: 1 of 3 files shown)


ekonstantinidis:
So does it work? Any issues? Also can you post a couple of screenshots? Will have a look on the PR later today(had a quick look - looks pretty good). If everything is ok, I think we can do the same for windows too!

caske33:
It works now, for me on ubuntu. I think there might be some issues on OS X now, especially this line: https://github.com/caske33/gitify/commit/faa7b98a755e5cfbbc567604067ab38171675f48#diff-7a9076d6d94e62c13d641aa71f19ae8eR101 (also an API that's only available on Linux)

Screenshots:
screenshot from 2015-05-31 19 33 46

I'm not sure if it's normal but when you click on the trayicon, you first have to click on a dropdown menu `Github Notifications on your menu bar` before the pop-up appears:

screenshot from 2015-05-31 19 34 28

main.js
```diff
@@ -93,6 +98,6 @@ app.on('ready', function(){
     app.quit();
   });
 
-  app.dock.hide();
+  //app.dock.hide();
```
ekonstantinidis: In order to avoid OSX issues this can be:

    if (app.dock) { app.dock.hide(); }

ekonstantinidis:
Can you also tell me what is the API that is only available to linux? Had a look on electron and we can check the platform with something like:

    if (process.platform === 'darwin') {
      // This is Mac OSX
    }

    if (process.platform === 'win32') {
      // This is windows
    }

    if (process.platform === 'linux') {
      // This is Linux
    }

> I'm not sure if it's normal but when you click on the trayicon, you first have to click on a dropdown menu Github Notifications on your menu bar before the pop-up appears (I'm unfortunately not able to take a screenshot of the dropdown menu)

This sounds wrong. On OSX when you click on the tray it pops up the browser window. The text `Github Notifications on your menu bar` is coming from the [tooltip](https://github.com/ekonstantinidis/gitify/blob/master/main.js#L97). What if you remove the tooltip text?

caske33:
https://github.com/atom/electron/blob/master/docs/api/tray.md#tray it seems that ubuntu doesn't support tray icons without submenus... :( (When I don't set the tooltip, I get an empty submenu).

I've added the platform check for the hiding of the dock.

(do you want me to squash commits or rather not? It's how you want it)

ekonstantinidis:
I think that the behaviour should be the same for all platforms. But as the docs say:

> So if you want to keep exact same behaviors on all platforms, you should not rely on clicked event and always attach a context menu to the tray icon.

I think we should have a further look at the options we have. I will try to install ubuntu and give it a go!

frmendes:
Any updates on this?

ekonstantinidis:
Hi @frmendes! I didn't have the chance to find an ubuntu machine(virtualbox is a pain) so there is no progress from my end. Shouldn't be hard to make it work though. Feel free to give it a go!

sergiovilar:
@caske33 what are your plans to distribute the version for ubuntu? Is some Ubuntu versions (14.04+) `libappindicator1` is required as dependency to Tray work properly. Maybe a .deb?

caske33:
@sergiovilar I've no experience whatsoever with distributing ubuntu versions. I'd love it if a .deb would appear, but I'm probably not going to make it, as i'm even more likely not to maintain it (which would be necessary).

ekonstantinidis:
@caske33 @sergiovilar Closing and moving this PR to #81 for further development! I've fixed all the conflicts so feel free to continue from there! (Your commits will remain on that PR)

caske33:
Ok, thank you!

PR #52, opened 2015-06-09
Settings - Participating Notifications

ekonstantinidis:


src/js/stores/notifications.js
```diff
@@ -22,9 +23,10 @@ var NotificationsStore = Reflux.createStore({
 
   onGetNotifications: function () {
     var self = this;
+    var participating = SettingsStore.getSettings().participating;
 
     apiRequests
-      .getAuth('https://api.github.com/notifications')
+      .getAuth('https://api.github.com/notifications?participating=' + participating)
```
JakeSidSmith: I would do the following just to make sure it's concatenated correctly.

```javascript
.getAuth('https://api.github.com/notifications?participating=' + participating ? 'true' : 'false')
```

JakeSidSmith:
:sparkles:

PR #56, opened 2015-06-12
Notifications / Sounds

ekonstantinidis:


src/js/stores/notifications.js
```diff
@@ -21,6 +22,35 @@ var NotificationsStore = Reflux.createStore({
...
+    var self = this;
+    var playSound = SettingsStore.getSettings().playSound;
+
+    if (!playSound) { return; }
+
+    // Check if notification is already in the store.
+    var isNew = false;
+    _.map(response, function (obj) {
+      if (!_.contains(self._previousNotifications, obj.id)) {
+        isNew = true;
```
JakeSidSmith: To keep this function pure you should use _.some...

```javascript
var isNew = _.some(response, function (obj) {
  return !_.contains(self._previousNotifications, obj.id);
});
```

src/js/stores/notifications.js
```diff
@@ -21,6 +22,35 @@ var NotificationsStore = Reflux.createStore({
...
+      }
+    });
+
+    // Play Sound.
+    if (isNew) {
+      if (playSound) {
+        var audio = new Audio('sounds/digi.wav');
```
JakeSidSmith: I would load this when the app first loads. :) Perhaps have a store for sounds / notifications.

src/js/stores/notifications.js
```diff
@@ -21,6 +22,35 @@ var NotificationsStore = Reflux.createStore({
...
+        audio.play();
+      }
+    }
+
+    // Now Reset the previousNotifications array.
+    self._previousNotifications = [];
```
JakeSidSmith:
```javascript 
self._previousNotifications = _.map(response, function (obj) {
  return obj.id;
});

```

ekonstantinidis: It should clean the 'previousNotifications' every time it receives a new payload so that's why..

src/js/stores/notifications.js
ekonstantinidis: As discussed will have a further look in cleaning it up on #57 .

PR #80, opened 2015-08-01
Update dist script & Remove client secret key

### LINKED ISSUE #75: App Secret in public repo
Maybe you don't want this in the repository since it is public? Excellent blog post about OAuth though ... much appreciated. 

https://github.com/ekonstantinidis/gitify/blob/master/src/js/components/login.js#L21

ekonstantinidis:
Closes #75 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "watch-js": "watchify -t babelify src/js/app.js -o build/js/app.js -v",
     "watch": "npm run build && grunt build && npm run watch-js & grunt watch",
     "start": "electron .",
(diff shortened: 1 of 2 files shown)


iKlsR:
Did this really change anything, the secrets are still in the repo and they can still be read even after the asar step. I guess ultimately it's a risk we have to make when using oauth clientside.

manosim:
Hello @iKlsR. This is quite an old PR - The keys that you see in the repository are used only for development - there are separate production keys. You can find more information in the README: https://github.com/manosim/gitify#development

I think we should investigate if there is something we can do to hide them after the asar step.

Issue #20, opened 2015-05-29
Improve Spaces / Multiple monitors behaviour 


richrace:
So I use multiple spaces.

Given `space_1` and `space_2`
I have Chrome open in `space_1`
When in `space_2` I click on Gitify thenI click on a Repo/PR/Issue it switches to `space_1`, perfect.
However, when I click back on Gitify, it then sends me to `space_2`.

Also, with multiple monitors (`monitor_1`, `monitor_2`).
1. When I open Gitify on `monitor_1`, click again on Gitify
2. Then open Gitify on `monitor_2`, it opens on `monitor_1`
3. Then clicking Gitify again to hide it, and again will show it on `monitor_2`

Again not sure if Electron or not.

ekonstantinidis:
I think that this has to do with Electron. I had a quick look in the Electron API and I couldn't find anything relevant to `getFocusedMonitor` on `TrayClick`. Will have to investigate further.

JTarasovic:
Would [BrowserWindow.setVisibleOnAllWorkspaces(true)](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#browserwindowsetvisibleonallworkspacesvisible) work?

ekonstantinidis:
@JTarasovic I had a look and even though this should help it didn't :cry: Trying to figure it out!

ekonstantinidis:
@richrace This has now been fixed at #82. Thanks! @JTarasovic It seems that what you suggested works great now :+1:

PR #82, opened 2015-08-11
Make visible menu window on all workspaces

rhysd:
Hi.
Thank you for such a great app.

In OS X, gitify is now bound to specific work space where gitify firstly started.  When I click the menu icon in other workspace, the window will be opened in the bound work space and current workspace is forced to be changed.

I used `BrowserWindow.setVisibleOnAllWorkspaces()` API to solve this issue.

https://github.com/atom/electron/blob/master/docs/api/browser-window.md#browserwindowsetvisibleonallworkspacesvisible

I think all menu bar applications should set this option to true.

### DIFF
--- a/main.js
+++ b/main.js
@@ -42,6 +42,7 @@ app.on('ready', function(){
     appIcon.window = new BrowserWindow(defaults);
     appIcon.window.loadUrl('file://' + __dirname + '/index.html');
     appIcon.window.on('blur', hideWindow);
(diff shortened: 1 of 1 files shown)


rhysd:
I think this PR can improve #20.

ekonstantinidis:
I've tried it before and I think it didn't work (maybe in older versions of Electron? who knows..) but now it works great! Thank you @rhysd! :rocket:

rhysd:
:sparkles::dog:

PR #131, opened 2015-12-08
Fix Window Positioning

### LINKED ISSUE #99: Clicking on OS X notification causes menu to appear far away from gitify icon
Clicked on OS X notification on the right hand side and menu appeared over to the left of the screen.

screen shot 2015-10-16 at 14 44 32

ekonstantinidis:
Fixed #99.


mrmlnc:
Works on Windows with taskbar in the lower position.
image

But it does not work with the top position of the taskbar.

ekonstantinidis:
@mrmlnc As discussed in Gitter will merge this one and we can work on top of master!

PR #144, opened 2015-12-30
fixed window's position on the second screen

sapjax:
If the user has two screens. click the tray will get wrong position.


ekonstantinidis:
Hi @sapjax! Yesterday I realised a new version that fixes those issues. Do you mind giving it a try and see how it goes? Thank you!

sapjax:
@ekonstantinidis 
window position is still not correct on the second screen.
15d0bbd8-a8f2-4f77-9e5d-52cecfa7db15

I found another issues: if i click  the "update" button many times, the update tip will alert many times.

PR #185, opened 2016-05-26
Additional clarity to Development section in README

TimMikeladze:
The client id and secret are defined in constants.js, not login.js (although they are consumed there).


ekonstantinidis:
Yes that's right! It changed when the app was re-written with redux! Thank you! 👍

PR #245, opened 2016-12-02
Clear interval when nav unmounts

JakeSidSmith:
This prevents a bug where after going on and offline for a while multiple intervals are started and causes a lot of requests, often resulting in rate-limiting.


JakeSidSmith:
@manosim, can you release this as a new patch version?

manosim:
Closing this one as the sidebar is nowadays always mounted (and I think we removed the online/offline functionality a few years ago 😅) - let's if this happens again.

PR #246, opened 2016-12-02
Link to watching

JakeSidSmith:


src/js/components/settings.js
```diff
@@ -98,6 +98,14 @@ export class SettingsPage extends React.Component {
           </div>
         </div>
 
+        <div className="row setting">
+          <div className="col-xs-12">
+            <a href="https://github.com/watching" target="_blank">
```
afonsojramos, line 103: If you check out the rest of the repo you'll see that we're using `shell.openExternal(url);` for opening external URLs.

PR #322, opened 2017-10-26
Close window upon clicking notification (setting)

ericboehs:
This pull request provides a new option to close the window upon clicking a notification in the list of notifications. This is handy for this using Gitify in "Tray mode".


manosim:
Closing this one as this should be the default functionality with v3(we are using menubar) - let's see if that's still the case!

PR #333, opened 2018-03-04
Allow application to be visible in fullscreen mode

mavimo:
This will allow to place application in fullscreen mode.

When the application is in full screen mode should be placed on the sidebar, like:

image


manosim:
Thank you for your PR @mavimo - Will have to close this one as the app has moved back to being attached to the tray icon in v3.

PR #379, opened 2020-02-15
Preferences: "showDockIcon"

### LINKED ISSUE #373: Option to display in dock only
I have so many things in my menu bar already, many things which can only exist in my menu bar. With the 3.0 upgrade, Gitify now looks to be menu-bar-only. In the screenshot below, Gitify is running, but I can't access it because the menu bar is full of other things:

image

I valued having Gitify _only_ in my dock, since it would never get hidden by other content, and I'd always see if something was pending. Now, unfortunately, it's unusable.

Should I downgrade to pre-3.0? Thanks for your help!

manosim:
Closes #373 .


afonsojramos:
Out of date PR.

PR #398, opened 2020-04-26
fix: add 'ci_activity' information to UI

codebytere:
This PR adds the contextual information for GitHub Actions workflows runs in the main UI, so that it no longer shows up as 'Unknown'. I picked the [sync octicon](https://octicons-git-v2.primer.now.sh/octicons/sync-16) for this new type but i'm of course happy to change it if you have another preference.

Before:
Screen Shot 2020-04-25 at 10 49 29 PM

After:
Screen Shot 2020-04-25 at 10 56 57 PM


src/types/github.ts
```diff
@@ -102,7 +104,7 @@ export interface Owner {
 
 export interface Subject {
   title: string;
-  url: string;
-  latest_comment_url: string;
+  url?: string;
```
codebytere, line 107: The `url` is not yet returned during a `ci_activity` notification response, which is why i changed this - i hope i can get GitHub to fix that up soon but in the meantime let me know if you'd prefer i not change this type and i'll remove it.

manosim: I think that's absolutely fine - I am hoping that the GitHub GraphQL API will support notifications at some point so these types will go away.

manosim:
Looks great - thank you for contributing @codebytere - I will bump the version and create a new release soon 🎉

PR #403, opened 2020-05-02
Raise test coverage to 100%

philihp:
This raises the line-by-line test coverage from 98% to 100%. I tried to keep my edits limited to only the tests, with the exception of `src/js/store/configureStore.ts` having a block of code that looks like it exists for debugging/development.

### DIFF
--- a/src/js/store/configureStore.ts
+++ b/src/js/store/configureStore.ts
@@ -34,6 +34,7 @@ export default function configureStore() {
     storageMiddleware,
   ];
 
+  /* istanbul ignore next */
   if (isDev) {
     const { createLogger } = require('redux-logger');
     const logger = createLogger({ collapsed: true });
--- a/src/js/store/configureStore.ts
+++ b/src/js/store/configureStore.ts
@@ -12,8 +12,6 @@ import notificationsMiddlware from '../middleware/notifications';
 import rootReducer from '../reducers';
 import settingsMiddleware from '../middleware/settings';
 
-const isDev = false;
-
 export default function configureStore() {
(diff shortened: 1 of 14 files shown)


src/js/store/configureStore.ts
```diff
@@ -34,6 +34,7 @@ export default function configureStore() {
     storageMiddleware,
   ];
 
+  /* istanbul ignore next */
```
manosim, line 37: Might not be related to this PR - but shall we remove this block completely (along with the `isDev` variable)? I guess we can also remove `redux-logger` from `package.json`.

src/js/utils/__snapshots__/github-api.test.ts.snap
```diff
@@ -78,6 +78,13 @@ Object {
 `;
 
 exports[`./utils/github-api.ts should format the notification reason 12`] = `
+Object {
+  "description": undefined,
```
manosim, line 82: Just realised that this is broken in the source file - shall we update the actual file https://github.com/manosim/gitify/blob/master/src/js/utils/github-api.ts#L48 to use `DESCRIPTIONS['CI_ACTIVITY']` which will remove that `undefined`.

src/js/store/configureStore.ts
philihp: Awesome! Even better 😎, I'll make this change.

I didn't want to overstep and really tried to avoid changing non-test files, but if you'd be okay with refactors, I'd be happy to go deeper and propose some cleanup changes.

philihp:
:shipit:

Issue #402, opened 2020-04-29
UI for unsubscribing a PR notification?


junjizhi:
I found I often click on the notification to go to the PR and press unsub button. It would be nice if I can mute the same series of notification through Gitify.

Is there a plan to support this feature?

JakeSidSmith:
I'd never really thought about that before, but I often do the same. Would be real nice. 😊

codebytere:
Seems pretty straightforward to potentially do so via [this endpoint](https://developer.github.com/v3/activity/notifications/#delete-a-thread-subscription) - i'd be happy to work on this potentially if that'd be something that would be acceptable to the project?

manosim:
I think it's a great idea, I tend to do the same thing(click on a specific link/issue and then press "unsubscribe". I had a look yesterday and was wondering what would be the correct place for such a button from a UX point of view.

Could be before the "check" in light gray but that would take more space from the notification title? Any thoughts?

codebytere:
Perhaps a circle slash octicon immediately to the left of the check? Perhaps we could make the window a tad wider to avoid removing too much text, but something like this?

<img width="516" alt="Screen Shot 2020-05-03 at 10 53 30 AM" src="https://user-images.githubusercontent.com/2036040/80921658-6cb37880-8d2c-11ea-91de-1fecf7b03169.png">

junjizhi:
I saw the small unsub button in Gitify app today. This is awesome. Thanks for the quick turn-around!

PR #407, opened 2020-05-03
feat: enable unsubscribing notifications

codebytere:
Closes https://github.com/manosim/gitify/issues/402.

When the new unsubscribe icon is clicked, all future notifications for a conversation are muted until one comments on the thread or gets an @mention.

Screen Shot 2020-05-04 at 4 32 02 PM

cc @manosim

### DIFF
--- a/src/js/actions/index.test.ts
+++ b/src/js/actions/index.test.ts
@@ -373,6 +373,81 @@ describe('actions/index.js', () => {
     });
   });
 
+  it('should unsubscribe from a notification thread with success - github.com', () => {
+    const id = 123;
+    const hostname = 'github.com';
+
+    // The unsubscribe endpoint call.
+    nock('https://api.github.com/')
+      .delete(`/notifications/threads/${id}/subscription`)
+      .reply(204);
+
+    // The mark read endpoint call.
+    nock('https://api.github.com/')
+      .patch(`/notifications/threads/${id}`)
+      .reply(200);
+
+    const expectedActions = [
+      { type: actions.UNSUBSCRIBE_NOTIFICATION.REQUEST },
+      {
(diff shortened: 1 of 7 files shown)


JakeSidSmith:
Can I suggest that we use the same icon as GitHub? 😉 
https://primer.style/octicons/mute

manosim:
I feel like it shouldn't be next to "✔" as it's not a primary action so what I did was to place it on the same line with the "reason" and "date" and it's visible only when you hover over a notification. This way we have more space more icons in the future (if needed). Thoughts on the gif? cc @codebytere @JakeSidSmith 

vJnAul77c4

JakeSidSmith:
I wouldn't have it hidden until hover, especially as it will also mark the notification as complete. I imagine people would be likely to couck it without meaning to and then struggle to find it again.

codebytere:
I'd agree with @JakeSidSmith - definitely better to make it a secondary action but perhaps keep it visible by default 🤔 

would something like this be fine, or would you prefer it right-aligned?

Screen Shot 2020-05-03 at 5 10 36 PM

manosim:
Maybe right aligned and always visible would be perfect if you agree? I think we will have to remove it from the list as well - like we do when pressing on the "check"?

codebytere:
@manosim i _believe_ this should be good to go now! Let me know if there are any other tests you'd like me to add, but i covered the basics based on what i saw was already there.

manosim:
Amazing - thank you for sorting this out @codebytere ! 🎉 
Will give it a couple of days and will cut a new release if nothing else comes up.

manosim:
@codebytere Can you try pushing an empty commit? It looks like GitHub Actions didn't pick up this PR 😢

Issue #417, opened 2020-06-19
Move to applications folder functionality on macOS


codebytere:
As it typically expected DX on macOS, an app should live in the Applications folder. As such, it'd be great if this app requested to move itself to the applications folder when it was first run, and then stored that preference on subsequent runs.

PR #418, opened 2020-06-19
feat: move to app folder on first run

codebytere:
Closes https://github.com/manosim/gitify/issues/417.

When gitify first opens, it will now request to move itself to the Applications folder on macOS. The user's choice will be stored for subsequent runs. If it's _already_ in the Applications folder, no prompts are shown.

cc @manosim I pulled the functionality into a new file but if you'd prefer I can put it into `main.js` or do something else entirely with it, up to you!

### DIFF
--- a/first-run.js
+++ b/first-run.js
@@ -1,0 +1,54 @@
+const { app, dialog } = require('electron');
+
+const fs = require('fs-extra');
+const path = require('path');
+
+async function onFirstRunMaybe() {
+  if (isFirstRun()) {
+    await promptMoveToApplicationsFolder();
+  }
+}
+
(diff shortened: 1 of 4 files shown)


codebytere:
@manosim yes - it [skips](https://github.com/manosim/gitify/pull/418/files#diff-8c91e6a6834d20a89d3ebf78dc6c2225R14) if we're not on darwin. `app.moveToApplicationsFolder` is macOS-only but this is primarily in that other platforms are less opinionated about application location and/or handle it at the system level, so there's not really a cross-platform expectation for this type of user pattern there I wouldn't say.

PR #419, opened 2020-06-19
chore: upgrade to Electron v9

codebytere:
This PR upgrades gitify to latest stable Electron v9.

This required few changes with the exception of changing the close button to a link. The reason behind this is that Chromium refreshed all their form controls recently (see [this blog post](https://blog.chromium.org/2020/03/updates-to-form-controls-and-focus.html)) which messed up the coloration of the button on the background:

Screen Shot 2020-06-19 at 10 53 03 AM

I figured that changing the button was easier than messing around with all the other background colors etc so that's what i went with. Happy to do something else if you'd prefer tho!

cc @manosim


manosim:
@codebytere Shall we also update the yarn lockfile? I think it's good to go apart from that 😄

codebytere:
Done!

manosim:
Amazing, thank you so much! Will cut a release branch to create a signed .dmg, test it for a while and then ship it 🚢

PR #420, opened 2020-06-23
Release 3.1.0

manosim:


manosim:
@codebytere Here's the link for 3.1.0 if you would like to give it a try/use it earlier https://github.com/manosim/gitify/releases/download/untagged-c9f398c43a02471bd72d/Gitify-3.1.0.dmg

PR #440, opened 2020-10-30
Optimize assets

danielgomezrico:
After running some optimization algorithms over the assets from https://imageoptim.com/mac the results are:

This optimizations does not change quality of the images.

image

> Saved 13 KB out of 323 KB. 35,6 % per file on average (up to 46,7 %)

Not so much but still more optimal 👀


danielgomezrico:
The images are now deprecated

Issue #458, opened 2020-12-15
Bug: Notification categories now fill the window incorrectly with v3.1.3


robert914:
When you get a project notification, the project name banner is filling the window now:

![2020-12-15 at 10 11 AM](https://user-images.githubusercontent.com/214286/102233415-3d6a0a00-3ebe-11eb-90f0-6fe131516eea.jpg)

This didn't use to happen prior to v3.1.3.

Please fix. Thanks!

crockpotveggies:
Confirmed, I am also experiencing this issue on MacOS 11.1.

mlostekk:
Same here

manosim:
Hi everyone, thank you so much for reporting this one -  this is now fixed and planning to release it later this week (hopefully tomorrow) along with some big changes as well (dark mode 😄 ). Apologies it took a bit longer, was trying to wrap up the migration from Redux to React Context which was a massive change and improvement going forward for Gitify. Stay tuned for version 4 🎉

PR #462, opened 2020-12-18
fix: oversized repository lines

studioromeo:
Heya! 👋 

There might be raised in a few places but this relates to #458. I noticed that we moved to tailwind recently so I think this was just a small bug from that refactor. 

What I've done here is set the flex of the repository item to `flex-initial` which is equivalent to `flex: 0 1 auto;` it lets the items shrink but not grow beyond their size. Hopefully this should do the job, it seems to work fine from my limited testing but you may want to give it a razz too!

This is my first contribution here so let me know if you want anything changed, thats no worries 🙂

### DIFF
--- a/src/js/components/__snapshots__/repository.test.tsx.snap
+++ b/src/js/components/__snapshots__/repository.test.tsx.snap
@@ -3,7 +3,7 @@
 exports[`components/repository.tsx should render itself & its children 1`] = `
 Array [
   <div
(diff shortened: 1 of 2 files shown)


manosim:
Hey @studioromeo, thank you so much for opening this PR! Gitify went through a major refactor this week (moved from redux to context etc) so this was fixed as part of that (#463). 🎉

PR #468, opened 2020-12-29
docs: Fix installation instructions in readme

kawarimidoll:
Fix error when installing using latest Homebrew. This is not a bug of this app, but readme needs to be fixed.

screenshot:

Calling brew cask install is disabled


manosim:
Was about to update that as well, thank you for the PR @kawarimidoll 😄

Issue #424, opened 2020-07-21
[Feature request] Github discussion support


yubozhao:
Thanks for the awesome tool!

As Github rolling out discussion feature to repos(https://github.blog/2020-05-06-new-from-satellite-2020-github-codespaces-github-discussions-securing-code-in-private-repositories-and-more), it would be very nice to also have discussion support by Gitify.

alavkx:
Seems Github discussions appear alongside other notifications (which is fine). Clicking on the discussion link dismissed the notification but does not open the link.

codebytere:
Unfortunately this is a bug in the GitHub notifications API - the `url` field in the notification payload is never populated for some newer types of Notifications, discussion included. I filed it with the team when I was still an employee there, but there isn't anything that can be done about it on this end; when GitHub fixes it it'll automatically start working here.

PR #487, opened 2021-03-18
Hotfix missing discussion items url

Yago:
I'm aware it's not super clean, but it does the trick until Github fix its REST API (see #424). If so, the first test will pass and this code will never reach out again and we will be able to remove it without second thought 😉

Again, it's just an idea for the mean time, feel free to say that it's a bad idea. I didn't update the test due to the very “temporary” aspect of this code, but I can do so if needed.

### DIFF
--- a/src/components/NotificationRow.tsx
+++ b/src/components/NotificationRow.tsx
@@ -38,16 +38,12 @@ export const NotificationRow: React.FC<IProps> = ({
         accounts.user?.id
       );
       shell.openExternal(url);
-    }
-
-    // TODO remove this future-legacy code when Github have pushed a fix (see #424)
-    if (
-      notification.subject.url === null &&
-      notification.subject.type === 'Discussion' &&
-      notification.repository.url
-    ) {
+    // For discussions, we can at least send users to the main discussions page.
+    } else if (notification.subject.type === 'Discussion') {
       const url = generateGitHubWebUrl(
-        `${notification.repository.url}/discussions`
+        `${notification.repository.url}/discussions`,
+        notification.id,
+        accounts.user?.id
       );
       shell.openExternal(url);
     }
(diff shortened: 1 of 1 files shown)


src/components/NotificationRow.tsx
```diff
@@ -35,6 +35,18 @@ export const NotificationRow: React.FC<IProps> = ({
       const url = generateGitHubWebUrl(notification.subject.url);
       shell.openExternal(url);
     }
+
```
codebytere, line 38: We know the first and third conditionals will always be true at this point of code execution, so we probably can clean this up a little bit:

```js
    // Some Notification types from GitHub are missing urls in their subjects.
    if (notification.subject.url) {
      const url = generateGitHubWebUrl(notification.subject.url);
      shell.openExternal(url);
    // For discussions, we can at least send users to the main discussions page.
    } else if (notification.subject.type === SubjectType.Discussion) {
      const url = generateGitHubWebUrl(`${notification.repository.url}/discussions`);
      shell.openExternal(url);
    }
```

codebytere (commented):
Thanks! Looking good, just a cleanup comment.

setchy:
Looking forward to this enhancement.  Thanks @Yago for contributing this PR

src/components/NotificationRow.tsx
Yago: Done 😉

setchy:
@manosim - do you think this enhancement is ready to merge and release?

Araxeus:
@manosim if this really fix the issue, why not merge this?

Its really annoying that discussions notifications straight up don't work

Araxeus:
we can get the actual Discussion url using GraphQL

I have created a repo to easily test this: https://github.com/Araxeus/github_discussion_notification_url

Any thoughts?

afonsojramos:
Will merge #538 over this one.

Issue #496, opened 2021-04-14
Bug: Does not change appearance with system


Ein-Tim:
## Setup:

- MacBook Air 2020 (M1)
- macOS Big Sur 11.2.3 
- Gitify v. 4.2.1

## Describe the issue:

When setting the appearance setting to "System", the appearance of Gitify never changes the appearance with the system. So even if the system changes to light mode, Gitify stays in dark mode (and the way around)

## Expected behavior:

Gitify should change its appearance with the appearance of the system.

PR #515, opened 2021-09-17
fix: dark mode not toggling with system

codebytere:
Closes https://github.com/manosim/gitify/issues/496.

Fixes an issue that could occur when the theme was set to `System`. In this case, if the user updated dark mode on their OS, the theme wouldn't update in the app unless it was quit and opened again. Fix this by using `nativeTheme` to listen for system theme changes and update the UI accordingly.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,4 +1,4 @@
-const { ipcMain, app } = require('electron');
+const { ipcMain, app, nativeTheme } = require('electron');
 const { menubar } = require('menubar');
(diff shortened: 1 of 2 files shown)


manosim (approved):
Looks good 🚀

Ein-Tim:
When will this fix be released?

Ein-Tim:
This PR did not successfully fix the issue. See #550

Issue #421, opened 2020-06-24
unsubscribe does not change subscription status for me


budparr:
V3.0.5, tested on v3.1.0 as well

Hi. I love the new unsubscribe button. It doesn't unsubscribe me from an issue though. After I click on the unsubscribe button, the issue disappears as I would expect, but if there's new activity on the issue it appears in Gitify again as if I did nothing. If I look at an issue I've unsubscribed from in Gitify, Github's button still indicates I can unsubscribe (indicating I'm still subscribed).

I've logged out of Gitify and re-authenticated, but it still have the issue.

I'm on Mac/Catalina 10.15.5

talyguryn:
same problem

gitify v4.1.0
macOS 10.15.7

JakeSidSmith:
Hmm, the request seems to just ignore the notification.

Request: 

<img width="585" alt="Screenshot 2021-01-08 at 09 00 23" src="https://user-images.githubusercontent.com/5850625/103995157-fa831580-518f-11eb-85fa-786fc44e035d.png">

<img width="139" alt="Screenshot 2021-01-08 at 08 57 22" src="https://user-images.githubusercontent.com/5850625/103994926-9d875f80-518f-11eb-8a1c-79023c2eeeba.png">

Response:

<img width="630" alt="Screenshot 2021-01-08 at 08 57 31" src="https://user-images.githubusercontent.com/5850625/103994928-9e1ff600-518f-11eb-9506-f0c583b46a51.png">

I think what we need is to delete the subscription: https://docs.github.com/en/free-pro-team@latest/rest/reference/activity#delete-a-thread-subscription

JakeSidSmith:
@manosim this needs be `DELETE`? https://github.com/manosim/gitify/blob/master/src/hooks/useNotifications.ts#L159

talyguryn:
@JakeSidSmith yea, at this moment "mute" button do the same as "mark as read".

looks like 'delete' method will fix this issue.

talyguryn:
@manosim any updates or ideas?

Juice805:
Same issue. Would be a huge QoL improvement to be able to unsubscribe

Juice805:
Looks they were trying to use [this endpoint](https://docs.github.com/en/rest/reference/activity#set-a-thread-subscription) but it should be `ignored`, rather than `ignore`

PR #517, opened 2021-09-27
fix: unsubscribe button ignoring thread subscription

Juice805:
Closes https://github.com/manosim/gitify/issues/421

The unsubscribe button isn't actually updating the subscription. From [here](https://docs.github.com/en/rest/reference/activity#set-a-thread-subscription) it appears the key should be `ignored` rather than `ignore` (even though [right below](https://docs.github.com/en/rest/reference/activity#delete-a-thread-subscription) mentions `ignore`).


Juice805:
This should be linked to #421

codebytere (approved):
hey thanks! Updated your PR body to close the issue when this merges.

Issue #527, opened 2021-11-10
Access Token Login not working with On-Prem GitHub Enterprise Servers


pbsureja:
Hello,

My company uses a non standard Github Enterprise API : https://github.company.com/api/v3 instead of standard api.github.company.com which Gitify seems to refer to when connecting. Is there a way for me to modify configuration use actual api instead of the one created by Gitify ?

ranma2913:
Can confirm this at my company too.  @pbsureja  I think you should update your Issue title though. It's not a non-standard api. In fact it IS standard for On-Prem Hosted GitHub Enterprise servers. See: https://docs.github.com/en/enterprise-server@3.1/rest/reference/oauth-authorizations

Maybe call the PR "Access Token Login not working with On-Prem GitHub Enterprise Servers"

When very closely reading the setup dialog we see "GitHub for Enterprise" which I'm guessing is a GitHub Cloud Service...
![image](https://user-images.githubusercontent.com/4295880/144090446-19cc8e21-8b8e-4fee-bfe9-5dadf68c2df9.png)

ranma2913:
A way which seems to work is if you use Oauth App Method:
1. Create new OauthApp:
![image](https://user-images.githubusercontent.com/4295880/144091827-344577c5-3b14-4a32-abba-35edaa412e52.png)
2. Doesn't really matter what you configure here, but this is what I put:
![image](https://user-images.githubusercontent.com/4295880/144092406-9ada37d6-0ea3-4030-a604-d52910c49acc.png)
3. Create new ClientSecret

LunaticMuch:
Has not been any more traction on this? I experience the same issue. Github Enterprise is not usable

PR #533, opened 2021-12-23
fix: github enterprise login

Uanid:
Hello,

I saw #524, Also I tested with latest commits.
But Login to Github Enterprise Server with PAT not works. (See #527)
Because Github Etnerprise Server's API url pattern is `https://${hostname}/api/v3`.

## Changes
This PR deduplicates api-url generate codes.

## Github Enterprise API url reference
https://docs.github.com/en/enterprise-server@3.3/rest/reference/users
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ temp/
 
 dist/
 build/
+
+# Jetbrains IDEs
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,3 @@ temp/
 
 dist/
 build/
-
-# Jetbrains IDEs
-.idea/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -78,3 +78,6 @@ temp/
 
 dist/
 build/
+
+# Jetbrains IDEs
+.idea/
```
codebytere, line 83: Can you please remove this? I'd prefer we not introduce changes orthogonal to the primary fix in the same PR.

codebytere (commented):
Minor nit, otherwise good to go.

.gitignore
Uanid: Thank you. I reverted that changes.

codebytere (approved):

pbsureja:
Thank you @Uanid for fixing this and @codebytere for reviewing this 🙂 . Do we know whose approval from maintainers do we need here ?

Araxeus:
@pbsureja want to make the same PR in my fork? https://github.com/Araxeus/gitify

afonsojramos (approved):
Thanks for the PR!

PR #539, opened 2022-03-08
fix: notification appId for windows

### LINKED ISSUE #546: Electron.app.Gitify Notification
Instead of `Gitify`, the notification name is `electron.app.gitify`.
I think this is electron issue, BUT I think you can fix it following this stackoverflow answer
https://stackoverflow.com/a/64090072/12291570

Araxeus:
change
image
to
image

fix #546


src/utils/notifications.ts
```diff
@@ -69,7 +69,7 @@ export const raiseNativeNotification = (
 
   if (notifications.length === 1) {
     const notification = notifications[0];
-    title = `Gitify - ${notification.repository.full_name}`;
```
codebytere, line 72: I see how this would be better looking on Windows, but on macOS this is important to help identify the app. Could you please make this platform specific?

codebytere (changes requested):

src/utils/notifications.ts
Araxeus:
```js
title = `${process.platform !== 'win32' ? 'Gitify - ' : ''}${notification.repository.full_name}`;
```
would that be ok?

codebytere: yes - let's go with that for now!

Araxeus: done in [8869d69](https://github.com/manosim/gitify/pull/539/commits/8869d69aee8b20dab0666a6aef2daa6a6bbd5c82)

afonsojramos (approved):

PR #542, opened 2022-03-17
fix: add RepositoryInvitation icon

codebytere:
Fix missing repository invitation icon.

Before:
Screen Shot 2022-03-17 at 10 46 30 AM

After:


afonsojramos (approved):
Looks good! Thank you!

PR #553, opened 2022-09-23
feat: settings button toggle between settings and notifications

### LINKED ISSUE #552: Settings button should toggle between Settings and Notifications, and clicking “Refresh” when viewing Settings should show Notifications/hide Settings
Sorry for the mouthful of a subject line. I’ve just noticed two minor UX issues I’d personally like to see remedied.

- When on the Settings page, clicking the Settings button does not return to Notifications (you need to explicitly click the back arrow for this)
- When on the Settings page, clicking the Refresh button does not  return to Notifications (you need to explicitly click the back arrow and then click Refresh)

I don’t know how easy it would be to remedy these, or if anyone else would even find them useful or not. Just thought I would try and explain what I’m after. Here’s a screen recording that might help:

https://user-images.githubusercontent.com/12046008/191888328-b04f8a3c-2ab9-4b20-8ae8-be5d10b6d03a.mov

hughlilly:
Closes #552. Not sure if this is the best way to go about what I’m trying to do, so I’d appreciate it if someone with much more experience than me would check over my suggestions. Thanks!

p.s. I think I stuffed up the commits. I wanted to make some other changes in [3d88d66](https://github.com/manosim/gitify/pull/553/commits/3d88d66368a4a14f64c0e64b75a6fb2eec3e7800) on another branch but they got included here. I hope I’ve reverted them properly.


src/components/Sidebar.tsx
```diff
@@ -63,15 +63,24 @@ export const Sidebar: React.FC = () => {
...
               aria-label="Refresh Notifications"
             >
               <IconRefresh className="w-3.5 h-3.5" />
             </button>
 
             <button
               className={footerButtonClasses}
-              onClick={() => history.push('/settings')}
+              onClick={() => {
+                if (window.location.hash.includes('settings')) {
```
JakeSidSmith, line 78: To check the current location you should use `react-router-dom`'s...

```ts
const location = useLocation();
```

And then:

```ts
if (location.pathname.startsWith('/settings')) {
```

(I think)

Gu7z:
@hughlilly @codebytere 🆙

src/components/Sidebar.tsx
codebytere: @hughlilly did you try the above approach?

afonsojramos (approved):

afonsojramos:
Nice QoL PR 👌

afonsojramos (approved):

hughlilly:
> Nice QoL PR 👌

Thanks; what does “QoL” mean?

afonsojramos:
> > Nice QoL PR 👌
> 
> Thanks; what does “QoL” mean?

Quality of Life 😅

PR #554, opened 2022-09-23
Change casing and phrasing of `onClickMarkAsRead` field

hughlilly:
Changes the casing and phrasing of the "On Click, Mark as Read" label to "Mark as read on click".

### DIFF
--- a/src/routes/Settings.tsx
+++ b/src/routes/Settings.tsx
@@ -93,7 +93,7 @@ export const SettingsRoute: React.FC = () => {
         />
         <FieldCheckbox
           name="onClickMarkAsRead"
-          label="On Click, Mark as Read"
+          label="Mark as read on click"
           checked={settings.markOnClick}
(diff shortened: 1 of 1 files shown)


src/routes/Settings.tsx
```diff
@@ -93,7 +93,7 @@ export const SettingsRoute: React.FC = () => {
         />
         <FieldCheckbox
           name="onClickMarkAsRead"
-          label="On Click, Mark as Read"
+          label="Mark as read on click"
```
codebytere, line 96: Should we keep this capitalized? i agree with making it a single no-comma sentence though

JakeSidSmith: I agree with sentence case to match the other fields. 😊 

Screen Shot 2022-09-23 at 17 57 00

afonsojramos (approved):

PR #555, opened 2022-09-23
Change dimensions of the login window

hughlilly:


JakeSidSmith:
For future reference, if you want to remove a commit, or adjust the last few commits I generally do:

```shell
git reset <commit-hash>
```

This will go back to the state of the commit that's hash you provide, with any changes you've made since then un-staged.

You can then discard any changes you didn't want to include and create your commits as usual.

To push to the remote you then use:

```shell
git push --force-with-lease
```

Force with lease allows you to overwrite what's on the remote, but won't overwrite anything that has been pushed by others (just to prevent undoing someone else's work in case they're working on the same branch).

JakeSidSmith:
Note: force pushes are dangerous, so make sure you're on the right branch.

PR #556, opened 2022-09-23
chore: change dimensions of login window

hughlilly:
This commit sets a more appropriate size for the login window (548 wide by 780 high).


codebytere:
@hughlilly is there an issue associated with this? why change it?

hughlilly:
It’s just not the right size for the content, IMHO. Currently it looks like this (800x600):
image

Here is 548x736:
image

I think it looks a lot better.

afonsojramos (approved):

PR #566, opened 2023-05-21
feat: switch to `pnpm` & update `runs-on` versions & packages

afonsojramos:
Some of the versions of the CI were no longer supported. Opted for going for `latest`, as latest in this context does not mean that it is not a mature platform.

`yarn` was also posing a blocker after upgrading the versions, so I decided to also take advantage to switch to pnpm, a community favorite that brings a few of advantages over yarn classic. This made it so that a few packages needed to be updated, as well as added, since `pnpm` is more strict with regards to dependencies.


.github/workflows/build-app.yml
```diff
@@ -4,106 +4,151 @@ on: [push]
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - name: Get yarn cache
-        id: yarn-cache
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Install Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
bmulholland, line 15: IMHO it's better to use `node-version-file` and reference the existing .nvmrc. That way there's only one source of truth for the node version. In this case, the `.nvmrc` is still set to v12 and should be updated too :). https://github.com/gitify-app/gitify/blob/ci/fix-ci-versions/.nvmrc

.github/workflows/build-app.yml
```diff
@@ -4,106 +4,151 @@ on: [push]
...
+
+      - uses: pnpm/action-setup@v2
+        name: Install pnpm
+        id: pnpm-install
+        with:
+          version: 8
```
bmulholland, line 21: It's not as strong an argument, but you could store the pnpm version in `package.json`'s `packageManager` field: https://github.com/pnpm/action-setup#version.

.github/workflows/build-app.yml
afonsojramos: Was not aware of `node-version-file`, but indeed good change, thanks!

.github/workflows/build-app.yml
afonsojramos: Sounds good to me! Less repeated code, even if it is in a `.yml`.

bmulholland (approved):
Nice work! It's always part pain, part satisfying to bring a project up to date.

You might also consider setting up renovate to automate the package updates. I'd suggest dependabot but they don't support pnpm, which is why I mention it in this context :)

afonsojramos:
@bmulholland I actually use depfu in another repository, but yeah, renovate, other than dependabot, is the industry standard. I'll try to get that set up.

bmulholland:
I use Depfu on a private repo, too. I don't think it supports pnpm though.

If you want to install Renovate, I've done it a few times. Here's the quick way to get it installed:
1. Go to https://github.com/apps/renovate and install it for the repo
2. [It will auto-create a PR "Configure Renovate"](https://docs.renovatebot.com/getting-started/installing-onboarding/#repository-onboarding) with a suggested config
3. I'd suggest just merging the PR's defaults. It's easiest to just see the resulting upgrade PRs, form an opinion on what you like/don't like, and tweak the config based on that.

afonsojramos:
I would have done it already, but unfortunately I only have write access for now, not admin access. Hopefully this can change in the future as having just one admin can be a blocker too often.

bmulholland:
Oh, yeah, well that's that then :)

PR #606, opened 2023-09-25
chore: add `repo` scope to token login

afonsojramos:
Add `repo` scope to the token creation link to conform with the Login With An Access Token page.
image

The `repo` scope has been added in https://github.com/gitify-app/gitify/pull/538


bmulholland:
What's this needed for?

afonsojramos:
@bmulholland added the details to the description.

bmulholland:
Sounds good! Looks like the change was clobbered, though?

afonsojramos:
> Sounds good! Looks like the change was clobbered, though?

I suppose that users can still remove the `repo` token if they're not comfortable with giving it, but for the default behaviour I suppose that we should aim to get maximum compat.

bmulholland:
I don’t think we’re talking about the same thing… I was saying that the diff is currently zero lines of code :)

afonsojramos:
Must have gotten mangled by the merge, sorry about that!

bmulholland (approved):

PR #644, opened 2023-10-03
feat: automate cask bump

afonsojramos:
I was a bit wary of using an external dependency here, but it makes it easier than dealing with the CLI directly, and since this is just a wrapper we should be fine.


bmulholland (approved):
Personally I'm all for wrappers :)

PR #645, opened 2023-10-03
fix: remove `outputFileSync` usage

afonsojramos:
I'm not sure where this came from since `outputFileSync` comes from `fs-extra`... I guess it always failed?


bmulholland (approved):
Nice catch

PR #648, opened 2023-10-03
chore(deps): Remove history dependency

bmulholland:
This was supposed to be removed as part of the react-router v6 upgrade (which moves history to a direct dependency), but it was clobbered by renovate

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+* @afonsojramos
+* @bmulholland
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,6 +1,10 @@
 name: Run Tests
 
-on: [push, pull_request]
+on: 
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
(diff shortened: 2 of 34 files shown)


bmulholland:
Current dependencies on/for this PR:
* main
  * **PR #648** <a href="https://app.graphite.dev/github/pr/gitify-app/gitify/648" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/gitify-app/gitify/648?utm_source=stack-comment).

bmulholland:
Ah I guess the tests have that dependency still... WDYT @afonsojramos, should we swap the tests and remove the direct dependency? Or keep the dev dependency and just upgrade the package?

afonsojramos:
> Ah I guess the tests have that dependency still... WDYT @afonsojramos, should we swap the tests and remove the direct dependency? Or keep the dev dependency and just upgrade the package?

If there is an alternative to it, I think that we can/should remove it!

bmulholland:
Wow, nice work. It's less code, too!

afonsojramos (approved):

PR #659, opened 2023-10-07
chore(CI): don't run tests twice per PR

### LINKED ISSUE #646: Duplicate tests running for every PR
Probably just a Github Actions configuration change needed?

Screenshot 2023-10-03 at 11 17 14

JoseAlberola:
This change resolves #646 by modifying the .github/workflows/run-test file to just execute the test on the pull request not in the push. Before tests were executed twice. 
Signed-off-by: JoseAlberola

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,6 +1,10 @@
 name: Run Tests
 
-on: [pull_request]
+on: 
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   run-unit-tests:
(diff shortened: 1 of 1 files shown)


bmulholland:
Thanks @JoseAlberola! I really appreciate you helping out with tasks :)

This change has a bit of collateral damage: it wouldn't run tests on the main branch. We need those so we can ensure the build is still working. Could you update the config so that still works?

bmulholland (changes requested):
As above

bmulholland (approved):

JoseAlberola:
You are welcome! I hope this configuration works for you :)

Issue #624, opened 2023-09-28
No New notifications after update to 4.4


eMerzh:
### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### 📝 Description

Hey 👋 

thanks for updating this app 🚀  

it seems that me and a few others have issues to retrieve new notifications.

i relogged, restart.. doesn't seems to help

### 🪜 Steps To Reproduce

1. start app
2. get notification
3. nothing happens

### ℹ System Info

```Markdown
Mac 13.5.2 (22G91) 
latest 4.4.0
```

afonsojramos:
Do you have notifications enabled for the app on MacOS?

eMerzh:
oh sorry that wasn't clear ... i meant github notifications 😬 ....
funnyliy enough i got at notif for this one .... now that i try to redo that i can't... 

if you want a screenshot :

<img width="795" alt="Capture d’écran 2023-09-28 à 20 37 45" src="https://github.com/gitify-app/gitify/assets/177003/4a0b4a48-869a-4577-ba7d-b65d94760938">

AlexanderHieser:
I have the same problem. Going back to v4.3.1 solved the issue.

eMerzh:
@afonsojramos hum ... it seems to be related to colors indeed 
(but i like colors 🥺)

ftr I used the token to login...

afonsojramos:
I think that for the colors it may require the `repo` scope, which the previous version did not have. The `repo` scope was introduced here: #538 
I'll try to take a look if there is something that can be done.

dberuben:
same issue

cleardemon:
H there, just wanted to confirm that either disabling colours or updating the token with `repo` scope makes notifications appear in Gitify again in 4.4.

ponkio-o:
Hi, I'm having the same problem with Github Enterprise Server. I tried `repo` scope permission granted but it didn't show up.
Does anyone else have a similar problem??

bmulholland:
@rbclark Yes, you got it, see https://github.com/gitify-app/gitify/issues/624#issuecomment-1741007292

Unfortunately we don't have a flow to request updated scopes, and doing that would be a bunch of work. We could add a fallback to not use colors, though that also wouldn't be great.

rbclark:
[...]

PR #654, opened 2023-10-04
fix(auth): Support 2FA via browser window auth

bmulholland:
I'm working to remove the usage of electron remote, and Auth windows are the last one. They open the auth in a new window. Since Github has started enforcing 2FA for almost everyone, the electron-window auth approach isn't functional for most. It therefore doesn't make sense to spend work maintaining the existing broken auth; instead, we can switch to a new one.

This change moves auth to open in the default browser. When Auth is done, it will redirect to `gitify://oauth-callback`, which the app will register and monitor, finishing the auth process when it's done.

At this point, I've got an initial approach started, but there's still lots to do. Help is welcomed.

There are several pieces left:
- [ ] Fix the start of the auth process
  - [ ] Allow the `gitify://oauth-callback` redirect in the Github app settings -- do we have access to that?
  - [ ] Fix the initialization of it -- all the types etc are broken, and the browser probably doesn't even open
- [ ] Catch and handle the completion of the auth
  - [ ] Communicate the auth from the main process to the App context, which started the Auth
  - [ ] Error handling
- [ ] Testing
  - [ ] Figure out automated testing approach, if possible
  - [ ] Manually test
- [ ] Documentation
  - [ ] Per https://github.com/gitify-app/gitify/issues/395, document the changed redirect URI

Fixes https://github.com/gitify-app/gitify/issues/561
Fixes https://github.com/gitify-app/gitify/issues/650
Fixes https://github.com/gitify-app/gitify/issues/429
Fixes https://github.com/gitify-app/gitify/issues/485
Closes https://github.com/gitify-app/gitify/issues/395

PR #660, opened 2023-10-09
fix(notifs): Default colors off

bmulholland:
Per https://github.com/gitify-app/gitify/issues/624, enabling colors for existing users is breaking notifications for those without the repo scope.

As a quick stopgap, this PR defaults colors to off and flags that re-auth is required.

Following this, we can add an error check that shows a message to the user when their request fails because they don't have the necessary scope. Alternatively, as #624 suggests, we can just log the user out in this case. When we've done that, we can re-enable colors as a default. I don't have time for this better fix today, so pushing this.

I recommend we release this as 4.4.1 to fix the immediate problem. WDYT @afonsojramos ?

### DIFF
--- a/src/context/App.test.tsx
+++ b/src/context/App.test.tsx
@@ -264,7 +264,7 @@ describe('context/App.tsx', () => {
         participating: true,
         playSound: true,
         showNotifications: true,
-        colors: true,
+        colors: false,
       },
     );
   });
@@ -302,7 +302,7 @@ describe('context/App.tsx', () => {
         participating: false,
         playSound: true,
         showNotifications: true,
-        colors: true,
+        colors: false,
       },
     );
   });
--- a/src/context/App.tsx
+++ b/src/context/App.tsx
@@ -37,7 +37,7 @@ export const defaultSettings: SettingsState = {
   markOnClick: false,
   openAtStartup: false,
   appearance: Appearance.SYSTEM,
-  colors: true,
+  colors: false,
 };
 
 interface AppContextState {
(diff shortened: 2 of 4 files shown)


bmulholland:
Current dependencies on/for this PR:
* main
  * **PR #660** <a href="https://app.graphite.dev/github/pr/gitify-app/gitify/660" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/gitify-app/gitify/660?utm_source=stack-comment).

bmulholland:
I have to go for today -- if anyone is able to fix the tests, it should be simple enough and will help us get this out faster.

afonsojramos:
I suppose that this is fine, but I'd prefer to simply fix the auth issue. Also, isn't #654 going to affect this?

bmulholland:
>  fix the auth issue

Which issue, exactly?

The problem, I think, is that currently-issued credential don't have repo scope. I think that's independent of #654, right?

afonsojramos:
Well, yes, but that PR will probably re-prompt everyone to sign in again and fix the issue by itself, right?

afonsojramos (approved):

afonsojramos:
What I am most confused about is how the normal login, not through the token, picks up the `repo` scope... Need to look into it

afonsojramos:
Just looked into it... 🤦 It requests it in the URL! We might as well simply reprompt for a login if they don't have the repo scope, right?

bmulholland:
Yeah I think that's the best solution for now. See https://github.com/gitify-app/gitify/issues/624#issuecomment-1751146302

Want to take a stab at that instead?

PR #666, opened 2023-10-10
chore: set renovate to run monthly

afonsojramos:


renovate.json
```diff
@@ -3,5 +3,6 @@
   "extends": ["config:base"],
   "labels": ["dependency"],
   "prConcurrentLimit": 5,
-  "rangeStrategy": "pin"
+  "rangeStrategy": "pin",
+  "schedule": ["monthly"]
```
bmulholland, line 7: Are you sure this is a valid schedule value? I think schedule:monthly is actually a preset, set above in "extends". I could be wrong though.

I've previously used https://github.com/marketplace/actions/validate-renovate-configuration-with-renovate-config-validator which would confirm this for us. Worth adding?

afonsojramos: From my understanding, yeah, it is valid. The CLI also approved it, just applied one suggestion! Will add that to the validations as well I suppose!

bmulholland: Nice, good stuff. Thanks for adding and checking!

bmulholland (approved):

PR #669, opened 2023-10-16
fix: use enterprise token if hostname of notification is enterprise server 

p1ass:
## Context

Thanks to many contributions released in https://github.com/gitify-app/gitify/releases/tag/v4.4.0, almost issues related to GitHub Enterprise Server have been fixed.
However, an issue about token usage still remains.

## Summrary

* Fix to use an enterprise token if the notification's hostname is enterprise server's name.
	* Current implementation used a github.com token for everything, so some API calls failed with a 401 error.


afonsojramos (approved):

afonsojramos:
Thanks for the PR!

PR #671, opened 2023-10-16
refactor: replace substr with substring

setchy:


src/utils/helpers.ts
```diff
@@ -52,7 +52,7 @@ export function generateGitHubWebUrl(
 
   if (newUrl.indexOf('/releases/') !== -1) {
     newUrl = newUrl.replace('/repos', '');
-    newUrl = newUrl.substr(0, newUrl.lastIndexOf('/'));
+    newUrl = newUrl.substring(0, newUrl.lastIndexOf('/'));
```
afonsojramos, line 55: As far as I am aware the second argument changes from being `length` for `substr()`, to `end` for `substring()`, as such, does this code still work? Does our tests cover this?

setchy: Should still work.  lastIndexOf(/) returns a length and index.  tests still pass

afonsojramos: What does the change do then, if both work? What is its purpose?

PR #673, opened 2023-10-17
feat: check for github scopes

afonsojramos:
Alternative to #660, also attempting to fix https://github.com/gitify-app/gitify/issues/624, since enabling colors for existing users is breaking notifications for those without the repo scope.

This PR notifies the user every time that he doesn't have the required scopes and disables colors in case he doesn't. Opted by not logging out to not be too annoying.

Warning: Implementation might not be the best.

### DIFF
--- a/src/context/App.tsx
+++ b/src/context/App.tsx
@@ -37,7 +37,7 @@ export const defaultSettings: SettingsState = {
   markOnClick: false,
   openAtStartup: false,
   appearance: Appearance.SYSTEM,
-  colors: true,
+  colors: false,
 };
 
 interface AppContextState {
--- a/src/context/App.tsx
+++ b/src/context/App.tsx
@@ -1,11 +1,13 @@
 import React, {
-  useState,
   createContext,
   useCallback,
   useEffect,
   useMemo,
+  useState,
 } from 'react';
 
+import { useInterval } from '../hooks/useInterval';
+import { useNotifications } from '../hooks/useNotifications';
 import {
   AccountNotifications,
   Appearance,
@@ -15,14 +17,12 @@ import {
   SettingsState,
 } from '../types';
 import { apiRequestAuth } from '../utils/api-requests';
-import { addAccount, authGitHub, getToken, getUserData } from '../utils/auth';
-import { clearState, loadState, saveState } from '../utils/storage';
 import { setAppearance } from '../utils/appearance';
+import { addAccount, authGitHub, getToken, getUserData } from '../utils/auth';
 import { setAutoLaunch } from '../utils/comms';
-import { useInterval } from '../hooks/useInterval';
-import { useNotifications } from '../hooks/useNotifications';
 import Constants from '../utils/constants';
 import { generateGitHubAPIUrl } from '../utils/helpers';
+import { clearState, loadState, saveState } from '../utils/storage';
 
 const defaultAccounts: AuthState = {
   token: null,
--- a/src/routes/Settings.test.tsx
+++ b/src/routes/Settings.test.tsx
@@ -1,27 +1,33 @@
+import { fireEvent, render } from '@testing-library/react';
 import React from 'react';
-import TestRenderer, { act } from 'react-test-renderer';
-import { render, fireEvent } from '@testing-library/react';
 import { MemoryRouter } from 'react-router-dom';
+import TestRenderer, { act } from 'react-test-renderer';
 
 const { ipcRenderer } = require('electron');
 
-import { SettingsRoute } from './Settings';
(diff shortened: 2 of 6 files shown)


afonsojramos:
A test to validate that it does not allow to disable colors needs to be added, but I need to fix this axios error beforehand

src/context/App.tsx
```diff
@@ -37,7 +37,7 @@ export const defaultSettings: SettingsState = {
   markOnClick: false,
   openAtStartup: false,
   appearance: Appearance.SYSTEM,
-  colors: true,
+  colors: false,
```
bmulholland, line 40: Maybe I'm missing something; I thought the intent of this PR was to keep colors enabled by default?

src/routes/Settings.tsx
```diff
@@ -28,6 +32,27 @@ export const SettingsRoute: React.FC = () => {
...
+    (async () => {
+      const response = await apiRequestAuth(
+        `${generateGitHubAPIUrl(Constants.DEFAULT_AUTH_OPTIONS.hostname)}`,
+        'GET',
+        accounts.token,
+      );
+
+      console.log(JSON.stringify(response.headers));
+      const missingScopes = Constants.AUTH_SCOPE.filter((scope) => {
+        return !response.headers['x-oauth-scopes'].includes(scope);
```
bmulholland, line 45: Nice, I was wondering if there's a way to check the granted scopes.

src/routes/Settings.tsx
```diff
@@ -28,6 +32,27 @@ export const SettingsRoute: React.FC = () => {
...
+      });
+      if (missingScopes.length > 0) {
+        new Notification('Gitify - Permissions', {
+          body:
+            'You need to grant all the permissions to use this app. Missing scopes: ' +
```
bmulholland, line 50: Should this include an instruction of what they should do next?

bmulholland: Ah I'd thought that you weren't logging them out? From "Opted by not logging out to not be too annoying."

afonsojramos: @bmulholland can you take another look? Now I'm not polluting the notifications, just disabling the option if the repo scope is missing. I'm also memoizing with the token as dependency so that we're not constantly checking for the scope.

src/routes/__snapshots__/Settings.test.tsx.snap
```diff
@@ -164,6 +170,11 @@ exports[`routes/Settings.tsx should render itself & its children 1`] = `
         <label
           className="font-medium text-gray-700 dark:text-gray-200"
           htmlFor="showOnlyParticipating"
+          style={
+            {
+              "textDecoration": undefined,
```
bmulholland, line 175: Why add this?

src/routes/Settings.test.tsx
```diff
@@ -250,4 +311,73 @@ describe('routes/Settings.tsx', () => {
...
+
+    await findByLabelText('Use GitHub-like state colors');
+
+    fireEvent.click(getByLabelText('Use GitHub-like state colors'));
+
+    expect(updateSetting).toHaveBeenCalledTimes(1);
+    expect(updateSetting).toHaveBeenCalledWith('colors', true);
+  });
+
+  it('should not be able to disable colors', async () => {
```
bmulholland, line 354: Ideally the `it` statement or context would communicate the circumstances under which enabling/disabling colors is allowed.

Also, I think this test might be intending to check that you cannot _enable_ colors?

Finally, I don't think the expectation is actually confirming that the option is disabled, right? Like, I think the expectation would also pass if the option is enabled.

src/routes/Settings.tsx
```diff
@@ -30,6 +40,21 @@ export const SettingsRoute: React.FC = () => {
...
 
+  useMemo(() => {
+    (async () => {
+      const response = await apiRequestAuth(
+        `${generateGitHubAPIUrl(Constants.DEFAULT_AUTH_OPTIONS.hostname)}`,
+        'GET',
+        accounts.token,
+      );
+
+      console.info("Token's scopes:", response.headers['x-oauth-scopes']);
```
bmulholland, line 51: Intended?

bmulholland (approved):
Changes look generally good, just the comments about testing

src/routes/Settings.tsx
afonsojramos: Yeah, it might be useful for people to know their tokens' scopes.

afonsojramos: Although, we don't make it that easy to access devtools... I'll remove it

PR #675, opened 2023-10-17
chore: remove `lodash` and `emojione`

afonsojramos:
Small experiment of removal of lodash and emojione, our biggest culprits in terms of bundle size.

Regarding lodash, it is a no brainer. It removes almost 600Kb and it doesn't affect readability that much, if at all.
Regarding emojione, I'm a bit torn, because we would be relying on the OS render of the emojis, which means that emojis will look different depending on where it is running. Which reduces consistency (?)

### Before
image

### After
image


bmulholland:
On the other hand, maybe OS emojis will make it feel more natural in that environment? Hard to say.

We could try it. Easy to roll back :)

src/components/AccountNotifications.tsx
```diff
@@ -14,10 +13,21 @@ interface IProps {
 export const AccountNotifications = (props: IProps) => {
   const { hostname, showAccountHostname, notifications } = props;
 
-  const groupedNotifications = _(notifications)
-    .groupBy((obj) => obj.repository.full_name)
-    .sortBy((_, key) => key)
-    .value();
+  const groupedNotifications = Object.values(
+    notifications.reduce(
```
bmulholland, line 17: FWIW, my eyes glaze over with this, where the previous code was understandable at a glance.

Any way to maintain the readability?

bmulholland: Not a blocker

PR #689, opened 2023-12-06
refactor: use brew CLI over custom action

afonsojramos:
Custom action was broken, so this PR replaces it with the official `bump-cask-pr` CLI, which can be seen working here: https://github.com/gitify-app/gitify/actions/runs/7108934623/job/19353060346


bmulholland (commented):
Nice :)

PR #690, opened 2023-12-06
deps: remove coveralls

afonsojramos:
It seems that we're only using the GH action, not the package, so we should be free to remove this.


bmulholland (approved):
Sure

Why draft?

afonsojramos:
I just wanted to make sure that nothing would break and forgot about it 😅

PR #692, opened 2023-12-06
deps: replace testing-library's `react-hooks` for `react`

afonsojramos:


bmulholland (commented):
Nice, this is a lot cleaner

afonsojramos:
There is still a lot to improve I think in the tests though!

bmulholland:
Yep -- but one step at a time :)

PR #693, opened 2023-12-11
chore: rename to `on-tag` and trigger web build

afonsojramos:
Create notice over pr so that there is no need for a PAT. In the future if we can get a fine grained token that would be better.

This is supposed to trigger a build so that Astro can fetch the new version, as this is still more performance than shipping Astro with something else to fetch the latest version on the client side as was tested in https://github.com/gitify-app/website/pull/33


bmulholland (commented):
LGTM

PR #713, opened 2024-01-19
refactor: remove now redundant `mark as read on click` functionality

### LINKED ISSUE #712: Remove "Mark as read on click" Setting, which has no effect now
See https://github.com/gitify-app/gitify/pull/706#issuecomment-1900315876

setchy:
Closes #712

### DIFF
--- a/src/components/NotificationRow.tsx
+++ b/src/components/NotificationRow.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useContext } from 'react';
 import { formatDistanceToNow, parseISO } from 'date-fns';
-import { MuteIcon } from '@primer/octicons-react';
+import { CheckIcon, MuteIcon } from '@primer/octicons-react';
 
 import {
   formatReason,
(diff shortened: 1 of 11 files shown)


src/components/NotificationRow.tsx
```diff
@@ -84,20 +84,6 @@ export const NotificationRow: React.FC<IProps> = ({
...
-          title="Mark as Read"
-          onClick={() => markNotification(notification.id, hostname)}
-        >
-          <CheckIcon
-            className="hover:text-green-500"
-            size={20}
-            aria-label="Mark as Read"
-          />
-        </button>
-      </div>
```
adufr, line 100: Why remove the `Mark as Read` button?

I agree that the setting to `Mark as Read on click` is now useless, but IMO we should keep the button as a way of clearing a notification without having to actually open it in Github!

setchy: Good catch - will correct @adufr

bmulholland (approved):

bmulholland:
That was fast! Thanks :)

afonsojramos (commented):
Still catching up on everything! Alright, looking good!

PR #745, opened 2024-02-06
feat: update/move "unsubscribe" icon

adufr:
## Context

As discussed in #706, because of the new "Mark as done" button, the "unsubscribe" button was weirdly positioned, so this PR fixes it by moving icons

I also updated the "unsubscribe" icon to match Github's one ([bell-slash](https://primer.style/foundations/icons/bell-slash-16/))

Before:
image

After:
image

## Discussion

I changed the order of the icons to something that felt more natural to me / something that looks more like Github's notifications interface (even though they don't provide a "mark notification as read" button).

I guess this is kind of subjective so if you see reasons to not change it, please let me know 😛


adufr:
Uh oh, idk what happened with my branch management but I was only expecting 2 commits here, not 10, I might have been doing something wrong

setchy:
Looks much cleaner @adufr - thank you

setchy (approved):

bmulholland (approved):
Yeah that's better -- thank you!

setchy:
@adufr - with this now merged, I've been using this `main` branch today to do some other testing.

What i noticed with the new UX re: the 3 icons, is the following
* previously, the mouse hit-spot for the `mark as read` icon was the full row height.  now, it's only a portion of the row height
* previously, all of the `mark as read` icons were vertically aligned to the right, both for repository and notification rows.  this made going through the list really easy.  now, there is some variance horizontally for the `mark as read` icons which 
Screenshot 2024-02-15 at 5 18 54 PM

i'd like to propose we consider
* placing the `mark as read` on the far right
* increasing it so that it's hit-spot is the full row height
* aligning vertically

adufr:
@setchy I haven't used it much since I'm only running it while working on it, otherwise I'm still on 4.6.1..

I agree on moving the "Mark as read" icon to the right, which will allow us to align the icon with the "Mark repository as read" icon (which will need to be a bit smaller)

And concerning the button height, I'm proposing to make it `h-full`, while still staying inside the `NotificationRow`'s padding, this way it's a bit easier to click on the button, but it also prevents you from accidentally clicking the button of the next/previous notification row

Here's what it would look like (I'm about to open the PR):
image

bmulholland:
Thank you both for the attention to detail! Both the testing and the fixes are invaluable for the app to remain polished and usable. Much appreciated.

afonsojramos:
Interesting analysis there, thanks for the keen eyes 👀

Issue #714, opened 2024-01-19
Feature request: add a new "Mark as done on click" setting


adufr:
### 📝 Provide a description of the new feature

This feature would be possible thanks to the newly added `Mark notification as done` API endpoint (see #500)

It would basically do the same thing than the old `Mark notification as read on click` setting which is now obsolete (#712)

### ➕ Additional Information

I'm willing to implement this, but I'd like my other PR #500 to be merged before as it will use shared code/functions (I might open a stacked PR..)

rpuncel:
This would be hugely helpful; GitIfy is super useful but the "Done" feature in GitHub Notification Center is critical for me.  Right now, I'll let Gitify make me aware there's a notification, but then still open the GitHub Notification center to actually read the notification.

adufr:
I'm closing this as the feature has been merged 🎉

fredrikaverpil:
Was there ever a release that included this feature?

adufr:
@fredrikaverpil not yet :(

setchy:
@fredrikaverpil @adufr - we need to fix this before our next release #https://github.com/gitify-app/gitify/issues/696

PR #746, opened 2024-02-06
feat: add "Mark as done on open" setting

adufr:
## Issues related

Implements #714 

## Context

This PR introduces a new setting which I called "Mark as done on open" (defaults to `false`).

It takes advantage of the new API to mark a notification as done, and replaces the old "Mark as read on click" setting (which was removed in #713 because it became obsolete).

The PR also includes test, I pretty much copy-pasted everything from #713 to implement both the feature and the test, so there's really nothing new 😉  

## Discussion

I decided to name the setting "Mark as done on **open**" instead of "Mark as done on **click**" because "click" felt too vague to me (click on what?)

This can be discussed.


setchy (approved):
lgtm

bmulholland (approved):
Thank you!

PR #748, opened 2024-02-09
feat: better notification type icons

adufr:
## Context

Make Gitify icons match Github icons

## Discussion

I had to pass the `state` parameter to `getNotificationTypeIcon` because I needed this context to find the correct icon. I made it optional because it felt more accurate.

I tried to find all different states/icons but I might be missing some; if so, I default to the original icon, so there shouldn't be any regression

## Before / After example

Before:
image

After:
image

### DIFF
--- a/src/components/NotificationRow.tsx
+++ b/src/components/NotificationRow.tsx
@@ -49,7 +49,10 @@ export const NotificationRow: React.FC<IProps> = ({
   };
 
   const reason = formatReason(notification.reason);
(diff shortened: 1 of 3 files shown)


afonsojramos (approved):
Pretty good and harmless PR, thanks! You touch on a good point though. I think we could probably have some regression testing with https://github.com/lost-pixel/lost-pixel

bmulholland (approved):

Issue #434, opened 2020-08-05
Make the popup window resizeable


a2sheppy:
Is there a reason the window isn't resizable? It would be nice if (a) it were resizable, and (b) the size were remembered across uses of the app, including across relaunches.

setchy:
> @setchy Any chance you can also reopen this one? Gitify is great but not really practical to use for me because of this particular size limit. (zooming out does not help as it would make the text too small).

Do you have a suggestion on how you'd like to see this supported? Is this something you'd be willing to implement if agreeable by the maintainers?

bmulholland:
@ssbarnea I want to back @setchy up here, and go one step further. Your comment at https://github.com/gitify-app/gitify/issues/1312#issuecomment-2733087493 isn't respectful to the work we do, for free, and so actually we're _less_ likely to reopen this at your request. Honestly I'd hesitate to even review a PR you'd submit, because we'd then have to maintain it afterward -- something you haven't demonstrated appreciation for.

Issue #433, opened 2020-08-05
Instead of 400 pixels, make the height relative to the display height


a2sheppy:
Currently, the popup for this menu extra on Mac comes up with a fixed height of 400 pixels. This is so very small and doesn't allow much of a window into the state of my GitHub activities. Instead, it should be based on the display height, taking the `workArea.height` value and subtracting a small amount to allow it to not just fill the height between menu bar and dock. I fiddled with this a bit and couldn't quite make it work, but didn't spend much time on it either (I tried some shortcuts that just didn't work). Figured I'd file this just in case others are interested in trying to make it happen.

JakeSidSmith:
It would be nice to be able to resize the height, rather than a fixed value. 🙂  Not sure of the limitation of electron windows though.

JakeSidSmith:
And I see you've just created #434 XD 👍

a2sheppy:
I've also changed `resizable` to `true` in my local fork and everything works fine, though I doubt the size is saved. :)

JakeSidSmith:
How is the positioning with resizable? I believe the reason this was not enabled originally was due to issues with the position of the window after resize (or something like that).

a2sheppy:
Seems fine to me. You can of course resize the window from the top, so it looks weird, but that doesn't mean it doesn't work just fine.

fdintino:
@JakeSidSmith I'm not so sure…it _was_ resizable before the 3.0 release. It got disabled in b8e8b169, but I'm not sure how intentional that was. Flipping `resizable: true` doesn't seem to have any ill effects.

PR #749, opened 2024-02-09
feat: resizable window

adufr:
## Related issues

- #433
- #434

## Context

Make the Gitify window resizable

## Discussion

I decided to make it so that it's resizable by default because IMO it's better to not have too many settings.

I used localStorage to persist the size of the window across restarts (note if you're testing it in dev, make sure you properly quit gitify using the close button instead of `cmd+c` the `pnpm start` process; otherwise the size sometimes doesn't get saved across restarts (idk why))

I had to make a small hack (`menubarApp.window.webContents.executeJavaScript`) to get access to the localStorage in the `main.js` file because it was undefined in `menubarApp.window`; if you have a cleaner way to do this please let me know!

### DIFF
--- a/main.js
+++ b/main.js
@@ -4,7 +4,7 @@ const { autoUpdater } = require('electron-updater');
 const { onFirstRunMaybe } = require('./first-run');
 const path = require('path');
 
-require('@electron/remote/main').initialize()
+require('@electron/remote/main').initialize();
 
 app.setAppUserModelId('com.electron.gitify');
 
@@ -12,7 +12,7 @@ const iconIdle = path.join(
   __dirname,
   'assets',
   'images',
-  'tray-idleTemplate.png'
+  'tray-idleTemplate.png',
 );
 const iconActive = path.join(__dirname, 'assets', 'images', 'tray-active.png');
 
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -52,7 +52,30 @@ const menubarApp = menubar({
 });
 
 menubarApp.on('ready', () => {
+  // Utility function to reposition the window so it stays in the tray center
+  function moveToTrayCenter() {
```
bmulholland, line 56: Why do the function definitions need to be in the ready scope? Couldn't they be at the top level scope? Would make for cleaner code that's easier to refactor.

afonsojramos (changes requested):
As is, I disagree with the goals of this PR. I have not seen a single menubar app that allows for resizing and for a good reason I think. Either small or big, all of them allow for resizing and are always the same size. If we opt for going the depending on the screen size I suppose might consider it a bit more, but even so, I am not so sure.

adufr:
@afonsojramos should I spend some time tinkering something that depends on the screen size or should we just close this PR and the related issues?

TBH I tried doing something because I saw 2 open issues about this and I also felt like the window was a bit too small to display multiples notifications, but I'm satisfied enough with the possibility to use `cmd+` and `cmd-` to zoom/unzoom inside the window...

PR #766, opened 2024-02-15
feat: support ci workflow notifications

setchy:
Closes https://github.com/gitify-app/gitify/issues/381 and https://github.com/gitify-app/gitify/issues/526


setchy:
Notification title parsing was required since the Notification Event payload doesn't contain any status, branch or workflow name details.  Similarly, I couldn't find an API to query for these given a notification id

setchy:
Although the [GET Check Suite[(https://docs.github.com/en/rest/checks/suites?apiVersion=2022-11-28#get-a-check-suite) API would be a perfect fit our needs, there is no joining identifier (that i can see) within the [List Notifications](https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#list-notifications-for-the-authenticated-user) response to be able to make it useful.

What we would have to do, to use the API, is [List Workflow Runs for a Repository](2024-02-15T13:54:0) and use date ranges to narrow the results, before parsing them.

setchy:
Split this into #819, #820 and #821

PR #768, opened 2024-02-15
fix: correct typos

setchy:
A few typo corrections


bmulholland:
In other projects I've set up the [typos](https://github.com/crate-ci/typos/) tool, and run it in CI. It's pretty simple:
```
      - name: Check spelling
        uses: crate-ci/typos@master
```
Is that worth doing on this project too?

setchy:
> In other projects I've set up the [typos](https://github.com/crate-ci/typos/) tool, and run it in CI. It's pretty simple:
> 
> ```
>       - name: Check spelling
>         uses: crate-ci/typos@master
> ```
> 
> Is that worth doing on this project too?

might be overkill for where we're at atm.  I only found these three.  if we continue to detect them, then yes, i'd agree

bmulholland (approved):

PR #784, opened 2024-02-18
feat: add "Unsubscribe from repository" button

adufr:
## Context

Add a "Unsubscribe from repository" button.
image

It uses the  ["Set a repository subscription" REST API](https://docs.github.com/en/rest/activity/watching?apiVersion=2022-11-28#set-a-repository-subscription), 

## Discussion

Two things:

- At first I was going to start implementing the "Mark repository as done" button (as it had already been discussed in #706, PR coming soon 😉), but having miss-aligned icons made me look into adding this feature first...
But I feel like it's really "heavy" on the UI, it's not so beautiful and it makes it harder to distinguish if the row is a Notification or a Repository...
What do you think?

- Following the "Unsubscribe from notification" implementation, I made it so that clicking on  Repository unsubscribe marks all of the repository's notifications as read. But this behaviour looks a bit odd to me; why automatically discard all those notifications? And why mark them as read instead of done? Perhaps it is better not to discard those notifications at all? If so, we would have to find a way to tell the user that the unsubscribe action has been successful...

### DIFF
--- a/src/components/Repository.test.tsx
+++ b/src/components/Repository.test.tsx
@@ -13,6 +13,7 @@ jest.mock('./NotificationRow', () => ({
 }));
 
 describe('components/Repository.tsx', () => {
+  const unsubscribeRepository = jest.fn();
   const markRepoNotifications = jest.fn();
 
   const props = {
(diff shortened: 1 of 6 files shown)


adufr:
TBH even though this feature could still be a nice to have; I'm not sure it's required.
Maybe the solution to the misalignment of the icons between `Repository` and `NotificationRow` could just be a space between the `Mark repository as done` and `Mark Repository as read` buttons:
image

Please tell me what you think so I can adjust my PRs :P

bmulholland:
I don't currently have mental space to wrap my head around everything in the description, but two thoughts that come to mind on a first pass:
1. There's an added twist for this one: it could be taken to mean "unsubscribe from this repository," which isn't what this is doing.
2. To solve the visual complexity of all these icons, we could show the icons only on hover. That would be discoverable (you naturally mouse over the items to open them).

adufr:
> There's an added twist for this one: it could be taken to mean "unsubscribe from this repository," which isn't what this is doing.

Yup, that's why I'm really not sure that it's a good idea to implement this 😬 

> To solve the visual complexity of all these icons, we could show the icons only on hover. That would be discoverable (you naturally mouse over the items to open them).

I thought about this too! If others agree with this we can try it?

setchy:
I like the idea of exploring showing the 3-notification buttons on mouse-over the row, which is similar to how GitHub behaves.

Re: `unsubscribe from repository`, i'm not sure this is needed.  GitHub when viewing notifications grouped-by repo only has the `mark as read` button

adufr:
I'm dropping this PR, I've been running a `main` build for a few days and the repository "done" button looks enough

PR #788, opened 2024-02-18
feat: mark repository as done

adufr:
## Context

Implements a "Mark repository as done" button, as discussed in #706

## Discussion

We had concerns about rate-limiting but [according to Github API's the limit is 5000 calls per hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-authenticated-users); so I don't think we need to worry about it.

I added a space between repository icons so they are aligned with `NotificationRow` icons, but it can be removed if https://github.com/gitify-app/gitify/pull/784 is implemented 

image

### DIFF
--- a/src/hooks/useNotifications.ts
+++ b/src/hooks/useNotifications.ts
@@ -43,6 +43,11 @@ interface NotificationsState {
     repoSlug: string,
     hostname: string,
   ) => Promise<void>;
+  markRepoNotificationsDone: (
+    accounts: AuthState,
+    repoSlug: string,
+    hostname: string,
+  ) => Promise<void>;
   isFetching: boolean;
   requestFailed: boolean;
 }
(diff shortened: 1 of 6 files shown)


src/hooks/useNotifications.ts
```diff
@@ -312,6 +317,47 @@ export const useNotifications = (colors: boolean): NotificationsState => {
...
+
+        if (accountIndex !== -1) {
+          const notificationsToRemove = notifications[
+            accountIndex
+          ].notifications.filter(
+            (notification) => notification.repository.full_name === repoSlug,
+          );
+
+          for (const notification of notificationsToRemove) {
+            await markNotificationDone(
```
bmulholland, line 337: Is it possible/worth it to run these in parallel?

I guess we'd probably want to batch them, and then that blows up in scope pretty quickly, so maybe not?

Just asking questions :)

bmulholland (approved):

src/hooks/useNotifications.ts
adufr: That's a good point, I think it's better to run them in parallel, I've updated my code to use a [await Promise.all()](https://github.com/adufr/gitify/blob/bb609b9b829494e9a1b15ecd315a541d74d1d960/src/hooks/useNotifications.ts#L336-L344) 🙂

It also helps with the animation when notifications are removed because they were removed in a weird order, now they all disappear at the same time

bmulholland (approved):

afonsojramos (approved):
Niice, thanks!

PR #793, opened 2024-02-19
refactor: simplify notification html url logic

setchy:
Fetch the most appropriate html url for the notification based off the notification API response.

Order is
* notification last_comment_url
* notification subject url
* discussions specific lookup (not available via the notifications api)
* default, repository url

Removes or simplifies a lot of string manipulation to generate the link, add in comment anchors and referrer ids. 

Note: tests still need to be reworked given the logic changes in helper.ts


src/utils/helpers.ts
```diff
@@ -154,49 +142,33 @@ export const getLatestDiscussionCommentId = (
...
-      generateGitHubWebUrl(
-        notification.subject.url,
-        notification.id,
-        accounts.user?.id,
-        latestCommentId ? '#issuecomment-' + latestCommentId : undefined,
-      ),
-    );
+  } else if (notification.subject.url) {
+    url = await getHtmlUrl(notification.subject.url, accounts.token);
+  } else if (notification.subject.type === 'Discussion') {
```
bmulholland, line 158: What would you think about putting the above conditionals under a conditional for `notification.subject.type`, too? Then it would be a clear hierarchy of: for a discussion, do (logic), for a comment, do (other logic).

setchy: Moved this to a swtich statement and added a comment explaining the special handling block

src/utils/helpers.ts
```diff
@@ -154,49 +142,38 @@ export const getLatestDiscussionCommentId = (
...
-        accounts.user?.id,
-        latestCommentId ? '#issuecomment-' + latestCommentId : undefined,
-      ),
-    );
+  } else if (notification.subject.url) {
+    url = await getHtmlUrl(notification.subject.url, accounts.token);
   }
+
+  // Perform any specific notification type handling (only required for a few special notification scenarios)
+  switch (notification.subject.type) {
```
bmulholland, line 161: For defensive programming, should have a default case that throws a clear error that the type isn't handled.

setchy: Is the new update better?  there's like a dozen ways these guards can be structured :p

setchy:
This PR is now ready for review @afonsojramos  @bmulholland.  

In addition to simplifying the html_url logic (no more string manipulation), this now has increased unit test coverage, and in doing so resolved an edge-base bug for Discussions url generation.

src/utils/helpers.ts
bmulholland: It is! Thanks for the cleanup :)

bmulholland (approved):

bmulholland:
Yep, LGTM!

PR #810, opened 2024-02-22
chore: remove `delayedHideAppIcon()`

adufr:
It looks like the bug has been resolved; I'm running MacOS and I've had no issues with `showDockIcon: false`.

I think we can simply drop this function (I added `showDockIcon: false` to make it more explicit, even though it was already the default value)

### DIFF
--- a/main.js
+++ b/main.js
@@ -4,7 +4,7 @@ const { autoUpdater } = require('electron-updater');
 const { onFirstRunMaybe } = require('./first-run');
 const path = require('path');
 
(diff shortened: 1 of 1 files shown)


setchy:
Nice. Any hints in a changelog at when this may have been resolved?

setchy (approved):

adufr:
> Nice. Any hints in a changelog at when this may have been resolved?

Nope, [nothing there](https://github.com/max-mapper/menubar/blob/master/CHANGELOG.md).

PR #819, opened 2024-02-23
feat: add support for `approval_requested` reason

setchy:
As per https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#about-notification-reasons

### DIFF
--- a/src/utils/github-api.ts
+++ b/src/utils/github-api.ts
@@ -24,21 +24,22 @@ import { CheckSuiteStatus, Reason, Subject } from '../typesGithub';
 
 // prettier-ignore
 const DESCRIPTIONS = {
-    ASSIGN: 'You were assigned to the issue.',
-    AUTHOR: 'You created the thread.',
-    COMMENT: 'You commented on the thread.',
-    INVITATION: 'You accepted an invitation to contribute to the repository.',
-    MANUAL: 'You subscribed to the thread (via an issue or pull request).',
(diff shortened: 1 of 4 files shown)


afonsojramos:
If we are adding all of these new notifications, should we have a more granular control of what notifications gitify fetches to the user?

src/utils/github-api.ts
```diff
@@ -24,21 +24,22 @@ import { CheckSuiteStatus, Reason, Subject } from '../typesGithub';
 
 // prettier-ignore
```
afonsojramos, line 25: Why are we ignoring prettier here?

src/utils/github-api.ts
```diff
@@ -47,6 +48,8 @@ export function formatReason(reason: Reason): {
 } {
   // prettier-ignore
```
afonsojramos, line 49: Same here

src/utils/github-api.ts
setchy: 🤷 pre-existing. Can do a sweep in a follow up PR

setchy: pushed a new commit which only adds the new description... not sure how the others got reformatted

setchy:
> If we are adding all of these new notifications, should we have a more granular control of what notifications gitify fetches to the user?
> Just pointing out that we should probably roll that feature out before we roll out a new version.

Perhaps... there is a feature request ticket already for filtering by reason...

Note, we aren't adding ANY new notifications actually, what I've been doing is ensuring that we have fewer Unknown notifications that cannot be clicked on.

afonsojramos (approved):

Issue #813, opened 2024-02-22
feature: support for all discussion types (open, closed, duplicate, outdated)


setchy:
### 📝 Provide a description of the new feature

Add support for the 4 types of `Discussions`
* Open - `CommentDiscussionIcon` - color:gray
* Closed - `DiscussionClosedIcon` - color: purple
* Duplicate - `DiscussionDuplicateIcon` - color:gray
* Outdated - `DiscussionOutdatedIcon` - color:gray

setchy:
There is no indicator in the List Notifications API response.  Would need fetch this another way

Issue #823, opened 2024-02-26
refactor: use octokit client for fetching data from REST and GraphQL endpoints


setchy:
### 📝 Provide a description of the new feature

Instead of using our own types and api clients, refactor to use https://github.com/octokit/octokit.js

0xdevalias:
Another potential alternative:

> And `react-query` was mentioned a couple of times ([1](https://github.com/gitify-app/gitify/issues/1124#issuecomment-2181598890), [2](https://github.com/gitify-app/gitify/issues/1124#issuecomment-2249125693)), which is something I have worked with before and is quite nice.
>
> _Originally posted by @0xdevalias in https://github.com/gitify-app/gitify/issues/2284#issuecomment-3408544762_

PR #827, opened 2024-02-26
feat: support discussion state types

setchy:
Closes: #813

Uses the GraphQL API to fetch the Discussion `stateReason`

Screenshot 2024-02-27 at 6 58 00 AM

### DIFF
--- a/src/hooks/useNotifications.test.ts
+++ b/src/hooks/useNotifications.test.ts
@@ -3,7 +3,10 @@ import axios from 'axios';
 import nock from 'nock';
 
 import { mockAccounts, mockSettings } from '../__mocks__/mock-state';
-import { mockedDiscussionNotifications, mockedUser } from '../__mocks__/mockedData';
+import {
+  mockedDiscussionNotifications,
+  mockedUser,
+} from '../__mocks__/mockedData';
 import { AuthState } from '../types';
 import { useNotifications } from './useNotifications';
 
--- a/src/hooks/useNotifications.test.ts
+++ b/src/hooks/useNotifications.test.ts
@@ -345,6 +345,17 @@ describe('hooks/useNotifications.ts', () => {
               type: 'Discussion',
             },
           },
+          {
+            id: 6,
+            updated_at: '2024-02-26T00:00:00Z',
+            repository: {
+              full_name: 'some/repo',
+            },
+            subject: {
+              title: 'This is a default discussion',
+              type: 'Discussion',
+            },
+          },
         ];
 
         nock('https://api.github.com')
@@ -376,6 +387,14 @@ describe('hooks/useNotifications.ts', () => {
                     node: {
                       title: 'This is an open discussion',
                       stateReason: null,
+                      viewerSubscription: 'SUBSCRIBED',
+                    },
+                  },
+                  {
+                    node: {
+                      title: 'This is an open discussion',
+                      stateReason: null,
+                      viewerSubscription: 'IGNORED',
                     },
                   },
                 ],
@@ -426,6 +445,20 @@ describe('hooks/useNotifications.ts', () => {
                 ],
               },
             },
+          })
+          .post('/graphql')
+          .reply(200, {
+            data: {
+              search: {
+                edges: [
+                  {
+                    node: {
+                      title: 'unknown search result',
+                    },
+                  },
+                ],
+              },
+            },
           });
 
         const { result } = renderHook(() => useNotifications(true));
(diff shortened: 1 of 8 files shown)


setchy:
In a separate PR, i think it would be wise to move all of this `state` enrichment logic out of `hooks/useNotifications` and into a separate file.

src/hooks/useNotifications.test.ts
```diff
@@ -266,6 +281,263 @@ describe('hooks/useNotifications.ts', () => {
...
       });
+
+      it('should fetch discussion notifications with success - with colors', async () => {
+        const accounts: AuthState = {
+          ...mockAccounts,
+          enterpriseAccounts: [],
+          user: mockedUser,
+        };
+
+        const notifications = [
```
bmulholland, line 292: I don't love having the stubbed data mid-test like this.

src/typesGithub.ts
```diff
@@ -145,22 +157,31 @@ export interface Owner {
 export interface Subject {
   title: string;
   url: string | null;
-  state?: StateType;
+  state?: StateType; // This is not in the GitHub API, but we add it to the type to make it easier to work with
```
bmulholland, line 160: Thanks for adding the comment! Gives some ideas about what we could improve in future, too -- e.g. wrapping subject into our own type.

src/hooks/useNotifications.ts
```diff
@@ -146,28 +141,47 @@ export const useNotifications = (colors: boolean): NotificationsState => {
...
-                              : (cardinalData.draft && 'draft') || 'open';
-
-                          return {
-                            ...notification,
-                            subject: {
-                              ...notification.subject,
-                              state,
-                            },
-                          };
+                          switch (notification.subject.type) {
```
bmulholland, line 144: Moving this code out, as you suggest, will make a big difference to the mind-boggling level of indentation here :)

src/utils/helpers.ts
```diff
@@ -139,6 +139,53 @@ async function getDiscussionUrl(
...
+    response?.data?.data?.search?.edges?.filter(
+      (edge) => edge.node.title === notification.subject.title,
+    ) || [];
+  if (edges.length > 1)
+    edges = edges.filter(
+      (edge) => edge.node.viewerSubscription === 'SUBSCRIBED',
+    );
+
+  if (edges[0]) {
+    if (edges[0].node.isAnswered) {
```
bmulholland, line 177: If a discussion is answered, is the first entry always the answered one?

src/utils/helpers.ts
```diff
@@ -139,6 +139,53 @@ async function getDiscussionUrl(
...
+                title
+                stateReason  
+                isAnswered
+              }
+            }
+          }
+        }
+      }`,
+    });
+  let edges =
```
bmulholland, line 167: I'm missing context here. This looks like it's doing manual filtering. Why? Is there no better way? What are the edge cases and/or tradeoffs? Should those answers be in the code somehow?

src/typesGithub.ts
setchy: I was thinking of the same.  Ideally, I'd like to refactor to use #823 and it's ootb types for GitHub, and then have Gitify types that are narrowed down to the handful of fields needed to _actually_ power the experience (eg: title, color, icon, type, state, timestamp, url) rather than the full raw GitHub response types

setchy: https://github.com/gitify-app/gitify/issues/828 for future us

src/hooks/useNotifications.ts
setchy: https://github.com/gitify-app/gitify/issues/829 for future us

src/hooks/useNotifications.ts
```diff
@@ -146,28 +141,47 @@ export const useNotifications = (colors: boolean): NotificationsState => {
...
+                            case 'PullRequest':
+                              const cardinalData = (
+                                await apiRequestAuth(
+                                  notification.subject.url,
+                                  'GET',
+                                  token,
+                                )
+                              ).data;
+
+                              const state =
```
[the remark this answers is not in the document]
setchy: https://github.com/gitify-app/gitify/issues/829

src/utils/helpers.ts
setchy: I agree there may be an opportunity to simplify the GraphQL response parsing.  For now, I've kept it as similar to the Discussion Comment URL lookup logic.  

I'd like to do a few things with the GraphQL-based queries in separate PRs
* #823
* use GraphQL query variables, instead of string building (susceptible to injection issues)
* bypass querying for `edges` -> `node` and use the shortcut convention `nodes`

src/hooks/useNotifications.test.ts
setchy: Great feedback.  I've moved it out

bmulholland (approved):
Good for now -- thanks for the followup ideas!

PR #848, opened 2024-03-01
refactor: use state to pass checksuite status around

setchy:


setchy:
coverage/coveralls job failure is due to how the percentage is calculated.

bmulholland (approved):
Nice cleanup

PR #849, opened 2024-03-03
feat: use regex to extract all checksuite attributes

setchy:
Improve match accuracy by using regex to extract all CheckSuite title attributes


setchy:
Coverage check failing due to how the percentage is calculated.  All lines are covered

afonsojramos (approved):
Nice find and nice improvement! Local testing looks good as well!

PR #850, opened 2024-03-03
feat: use regex to extract workflow-run attributes

setchy:
Use regex to match workflow run titles and extract attributes.  Will support https://github.com/gitify-app/gitify/issues/526

### DIFF
--- a/src/hooks/useNotifications.test.ts
+++ b/src/hooks/useNotifications.test.ts
@@ -237,6 +237,14 @@ describe('hooks/useNotifications.ts', () => {
               url: 'https://api.github.com/5',
             },
           },
+          {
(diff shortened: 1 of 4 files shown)


afonsojramos (approved):
I don't use GitHub to handle deployments, so I've actually never seen this type of notification, but code wise everything looks good!

setchy:
> I don't use GitHub to handle deployments, so I've actually never seen this type of notification, but code wise everything looks good!

Me either! I created it in my sandbox - https://github.com/setchy/gitify-notifications-sandbox/actions/workflows/waiting.yml

If/as we see other formats of WorkflowRun notifications, we can add support then.

PR #854, opened 2024-03-03
feat: support grouping notifications by date or repository

### LINKED ISSUE #836: feature: support grouping notifications by date or repository

https://github.com/notifications supports displaying notifications in two ways
* date
* repository

Currently gitify always will group by repository.  

This feature suggestion is to match the experience of the official GitHub web interface

### ➕ Additional Information

_No response_

AltinGruda:
Added a dropdown menu to group notifications by date or repository. UI changes were made based on how it looks on  https://github.com/notifications 

closes #836


setchy:
Thanks for the contribution @AltinGruda.  Functionally it seems to do the trick.

I'm wondering whether the toggle should either be in the `Settings` menu, or, on the left-hand-side, perhaps below the notification count bell.

@brandonweiss @afonsojramos - what do you both think?

AltinGruda:
I thought of putting it in the sidebar below the notification bell icon, with the icon shown below. I think that would be best, what do you think?
image

setchy:
> I thought of putting it in the sidebar below the notification bell icon, with the icon shown below. Let me know what you guys think.
> image
> 
> 

Let's see what that looks like

AltinGruda:
This is how it looks with the dropdown being below the notification icon. 

image
image

setchy:
Thanks @AltinGruda - certainly looks better on the left side. 

A suggestion I'd appreciate feedback on... What if instead of the drop-down we simply toggled between two icons, with suitable hover text?

afonsojramos:
> What if instead of the drop-down we simply toggled between two icons, with suitable hover text?

@setchy, I understand where you're coming from, but in filters world, that is not a common UX practice, and thus, I would stay away from it.

setchy:
> > What if instead of the drop-down we simply toggled between two icons, with suitable hover text?
> 
> @setchy, I understand where you're coming from, but in filters world, that is not a common UX practice, and thus, I would stay away from it.

I dont think what we're doing is filtering.  To me, it's really just a yes / no `group by repository`

afonsojramos:
Right, but in the future maybe we will implement other sorting/grouping capabilities within the same menu, hence why I think the way it is looks better

AltinGruda:
I've considered the suggestion to turn the filter into a toggle and created a preview (screenshots attached, have not pushed the changes). While toggles can offer a clear interface, I'm open to feedback on whether this aligns with our UX goals or if there are better alternatives.

image
image

AltinGruda:
Are we sticking with the dropdown or the toggle one?

afonsojramos:
I'd go with the dropdown, but the implementation needs fixing. Clicking outside of the dropdown should close the dropdown, not close the block the whole app.

AltinGruda:
I went ahead and implemented the functionality to close the dropdown when clicking outside of it. The issue where clicking outside the whole app closes seems to be on the main branch as well, so I dont think I caused that issue

setchy (changes requested):
@AltinGruda - this is looking pretty close.

I've noticed that when using the `Date` view, that the notifications aren't being correctly ordered by date (newest to oldest) which is the GitHub UI behavior - https://github.com/notifications

It seems that currently just hiding the repository header row....

Additionally, do you think it is possible to add something that indicates the currently selected filter option?

setchy:
replaced by https://github.com/gitify-app/gitify/pull/1273

PR #874, opened 2024-03-11
refactor(icons): replace logout icon with octicons prime version

setchy:
**Before**
Screenshot 2024-03-14 at 7 38 26 AM

**After**
Screenshot 2024-03-11 at 2 48 35 PM

PR #873, opened 2024-03-11
refactor(icons): replace add account icon with octicons prime version

setchy:
**Before**
Screenshot 2024-03-14 at 7 38 26 AM

**After**
Screenshot 2024-03-11 at 2 46 04 PM

PR #875, opened 2024-03-11
refactor(icons): replace cog icon with octicons prime version

setchy:
**Before**
Screenshot 2024-03-14 at 7 40 11 AM

**After**
Screenshot 2024-03-11 at 3 03 23 PM


afonsojramos (changes requested):
That icon feels soulless. I'd prefer to stay with the current one, at least with the current design of gitify.

afonsojramos:
Or maybe I just need to get used to it... I'm going to leave it simmering 🥘

setchy:
> Or maybe I just need to get used to it... I'm going to leave it simmering 🥘

Simmer away 😉.  I noticed that the octicons are much sharper.   

> But if we do it, we should join #873 #874 and #876 with this one. It doesn't make sense to have them in separate PRs imo

Personal preference perhaps... I'm very much in favor of small, independent, isolated PRs to assist with review speed and keep the line changed to the smallest releasable, particularly with refactoring PRs

afonsojramos:
Right, but in this case you are introducing icons that don't play well with each other if not done on one go.

I feel like if we switch to these icons we also need to completely revamp our UI to properly mimick GitHub's UI. As is, I still think ours are perfectly fine. If we do do these UI changes, then I think the icon changes make total sense

setchy:
In the interest of moving this forward, I've consolidated into a single PR https://github.com/gitify-app/gitify/pull/885

PR #885, opened 2024-03-15
refactor(icons): use standard octicons

setchy:
Replaces the refresh, settings, logout and add account icons with those from the octicons primer design system used extensively throughout our experience.

~Note: if we can find a suitable replacement for the Quit icon from https://primer.style/foundations/icons then we'll be 100% octicon powered.~

| | Light | Dark |
|-|-|-|
| **_Before_** | Screenshot 2024-03-15 at 7 21 30 AM |  Screenshot 2024-03-15 at 7 21 38 AM |
| **_After_**  |  Screenshot 2024-03-19 at 2 49 50 AM |  Screenshot 2024-03-19 at 2 49 38 AM |


bmulholland:
For the Quit icon, how about just an X? I know we use it elsewhere, but it's pretty standard for "close this app" and is in Octicons.

afonsojramos:
I'm glad that we're now grouping them in one PR, because I truly dislike these new icons. I understand the objective here, but they just look plain ugly and the app just doesn't look right with the change.

afonsojramos:
> I think that grouping these changes all in one PR truly gave us a new perspective in what the changes would entail.
> 
> From a UI cohesiveness perspective, they seem as if they don't belong, mostly the ones in the bottom bar. The size of the bottom bar also seems to have grown, but that seems like a simple mistake.
> 
> Food for thought: Maybe one alternative would be to go away with the bottom bar and adding a profile image with a dropdown with these three options, where the icons would be accompanied with a descriptive text. What do you think?

Rephrased for the sake of being more clear about the issues I see with the current change.

setchy:
> The size of the bottom bar also seems to have grown, but that seems like a simple mistake.

yes, that had stuck out to me too. adjust

setchy:
> Food for thought: Maybe one alternative would be to go away with the bottom bar and adding a profile image with a dropdown with these three options, where the icons would be accompanied with a descriptive text. What do you think?

i'm not a huge fan of putting these an additional click away (behind a dropdown)

setchy:
> From a UI cohesiveness perspective, they seem as if they don't belong, mostly the ones in the bottom bar.

Personal taste perhaps.  🤷  

imho, this switch to fully adopt Octicons / Primer Design System makes Gitify feel more consistent and unified 🙃

src/components/Sidebar.tsx
```diff
@@ -104,7 +103,7 @@ export const Sidebar: React.FC = () => {
             aria-label="Quit Gitify"
             onClick={quitApp}
           >
-            <IconQuit className="w-3.5 h-3.5" />
+            <XCircleFillIcon size={16} aria-label="Quit Gitify" />
```
afonsojramos, line 106:
```suggestion
            <XCircleIcon size={16} aria-label="Quit Gitify" />
```

src/components/Sidebar.tsx
```diff
@@ -1,13 +1,15 @@
-import { BellIcon } from '@primer/octicons-react';
+import {
+  BellIcon,
+  GearIcon,
+  SyncIcon,
+  XCircleFillIcon,
```
afonsojramos, line 5:
```suggestion
  XCircleIcon,
```

afonsojramos (changes requested):
If I am the minority here, I am open to conceding. But in that case, I think I'd switch for the no fill X, since we're not using Fill anywhere else 🧐

PR #901, opened 2024-03-19
fix: improve error handling

setchy:
Improve handling of attributes which may be null or responses that may be null.

I found some of these causing issues when testing locally


src/utils/api-requests.ts
```diff
@@ -4,7 +4,11 @@ export function apiRequest(
   url: string,
   method: Method,
   data = {},
-): AxiosPromise {
+): AxiosPromise | null {
+  if (!url && !method) {
+    return null;
```
bmulholland, line 9: Why not create & resolve a promise instead?

src/utils/api-requests.test.ts
```diff
@@ -6,7 +6,18 @@ jest.mock('axios');
 const url = 'https://example.com';
 const method = 'get';
 
-describe('apiRequest', () => {
+describe('utils/api-requests.ts', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('handles null parameters', async () => {
```
bmulholland, line 14: When would this happen? Is it a valid case, or should it be resolved at a higher level?

setchy: We do handle in most cases checking that the url is not null higher up in the calling business logic.

Perhaps this is overkill.

src/utils/helpers.test.ts
```diff
@@ -445,6 +445,34 @@ describe('utils/helpers.ts', () => {
           `https://github.com/manosim/notifications-test/discussions/612?${mockedNotificationReferrer}#discussioncomment-2300902`,
         );
       });
+
+      it('default to base discussions url when graphql query fails', async () => {
```
bmulholland, line 449: Nice one, I like it

src/utils/subject.ts
```diff
@@ -104,16 +104,12 @@ async function getGitifySubjectForDiscussion(
...
   }
 
-  let comments = discussion.comments.nodes;
-  let discussionUser = null;
-
-  if (comments?.length) {
-    discussionUser = getLatestDiscussionComment(comments)?.author.login;
-  }
+  const discussionUser = getLatestDiscussionComment(discussion.comments.nodes)
+    ?.author.login;
```
bmulholland, line 108: Does login need a `?` too? Not sure how JS treats this.

```suggestion
    ?.author?.login;
```

src/utils/subject.ts
```diff
@@ -202,10 +198,14 @@ function getWorkflowRunStatus(statusDisplayName: string): CheckSuiteStatus {
 async function getLatestCommentUser(
   notification: Notification,
   token: string,
-): Promise<User> {
+): Promise<User | null> {
+  if (!notification.subject.latest_comment_url) {
+    return null;
```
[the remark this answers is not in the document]
bmulholland: Promise.resolve I think

bmulholland (approved):
Some questions and suggestions, but overall this is a lot of hardening, thank you!

src/utils/subject.ts
setchy: I believe this is ok.  if we get an author object thats not null, then it _should_ contain the login attribute

PR #903, opened 2024-03-20
build: allow release workflow to be run manually

setchy:
Allow us to manually run the release workflow to confirm things like code-signing in future

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - 'release/v**'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - 'release/v**'
+  workflow_dispatch:
```
bmulholland, line 7: Worth a comment to explain this? it's not obvious

bmulholland (approved):

PR #910, opened 2024-03-20
fix(deps): downgrade to electron@19 to resolve macos code signing and startup crash

setchy:
Resolves https://github.com/gitify-app/gitify/issues/696

After exhaustive testing, electron@20 or higher causes the issue

Once merged, I'll mark the new Renovate electron@20 PR as requiring code changes to prevent macos startup crashes


bmulholland (approved):
Amazing work here!

PR #914, opened 2024-03-21
build(prettier): include html, json, md files

setchy:


bmulholland (commented):
I always love more linting :)

PR #924, opened 2024-03-25
refactor: make explicit the different user types

setchy:
Previously, a shared User type was being used to represent
* Gitify Auth User - persisted once after authentication
* Response object for Issues/PRs
* Enhanced Subject details for Gitify

This update makes the types separate and explicit


bmulholland (approved):
Really nice cleanup, always good to split up jumbled concepts, thank you!

afonsojramos (commented):
Nice cleanup there!

PR #929, opened 2024-03-27
feat(settings): ignore bot notifications

setchy:
Add setting to control whether notification from `Bot` accounts are shown or hidden.

Closes #591

PR #930, opened 2024-03-27
refactor(settings): adjust margins

setchy:
Adjust top and bottom padding/margins in preparation for #929 

**Before**
Screenshot 2024-03-27 at 6 10 58 AM

**After**
Screenshot 2024-03-27 at 6 29 49 AM


bmulholland:
Could use another header or two, if you're tweaking this screen :). The one saying "Appearance" and then nothing else is a bit strange when I stop to think about it.

bmulholland (approved):

PR #941, opened 2024-03-30
refactor: extract notification count into reusable function

setchy:


src/components/Sidebar.tsx
```diff
@@ -32,12 +33,7 @@ export const Sidebar: React.FC = () => {
     ipcRenderer.send('app-quit');
   }, []);
 
-  const notificationsCount = useMemo(() => {
-    return notifications.reduce(
-      (memo, account) => memo + account.notifications.length,
-      0,
-    );
-  }, [notifications]);
+  const notificationsCount = getNotificationCount(notifications);
```
afonsojramos, line 36: I suppose that the big difference here is that we are not memoizing anymore, but considering that notifications was getting updated very often, this won't make big change in terms of rendering.

afonsojramos (approved):

PR #947, opened 2024-04-02
feat: redesign settings ux

setchy:
Redesign the *Settings* screen so that it supports our growing configuration options (ie: #945).

Main changes include
* fixed header
* fixed footer
* scrollable settings section
* group settings into three distinct buckets (appearance, notification, system)
* rename `settings.appearance` -> `settings.theme` 

Screenshot 2024-04-02 at 9 38 39 AM
Screenshot 2024-04-02 at 9 38 45 AM

PR #948, opened 2024-04-02
refactor: clarify show notifications setting

setchy:
Clarify that `showNotifications` is for native system notifications.  Move to new **System** section


afonsojramos (approved):
Was a bit confused for a bit, but then I noticed #947

PR #952, opened 2024-04-02
refactor: temporarily disable oauth login method

setchy:
Until we have the opportunity to resolve the OAuth Login flow, let's disable it and promote PAT as a interim method 

Screenshot 2024-04-02 at 4 23 51 PM

### DIFF
--- a/src/routes/Login.tsx
+++ b/src/routes/Login.tsx
@@ -1,6 +1,6 @@
 const { ipcRenderer } = require('electron');
 
-import React, { useCallback, useContext, useEffect } from 'react';
+import React, { useContext, useEffect } from 'react';
 import { useNavigate } from 'react-router-dom';
 
 import { AppContext } from '../context/App';
@@ -8,7 +8,7 @@ import { Logo } from '../components/Logo';
 
 export const LoginRoute: React.FC = () => {
   const navigate = useNavigate();
-  const { isLoggedIn, login } = useContext(AppContext);
+  const { isLoggedIn } = useContext(AppContext);
 
   useEffect(() => {
     if (isLoggedIn) {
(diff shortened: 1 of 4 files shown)


src/routes/Login.tsx
```diff
@@ -36,13 +36,22 @@ export const LoginRoute: React.FC = () => {
         GitHub Notifications <br /> on your menu bar.
       </div>
 
-      <button
+      {/* <button
```
bmulholland, line 39: It's probably okay as-is, since we plan to get to this soon, but usually I prefer to add a comment explaining why a section is commented out.

For example, if we forget to remove a commented-out section, how will we understand why it's there (and that it can be removed) in six months?

src/routes/LoginWithToken.tsx
```diff
@@ -71,11 +71,11 @@ export const LoginWithToken: React.FC = () => {
               >
                 personal access tokens
               </a>{' '}
-              and create one with the{' '}
+              and create one with the three scopes{' '}
```
bmulholland, line 74: Any worry that "three" could become outdated?

bmulholland (approved):
Fast work! Loving all the attention this stuff is getting, thank you :)

Left a couple minor comments

src/routes/Login.tsx
setchy: Good callout.  I've added a `FIXME` comment and reference the currently known open GH Issues

src/routes/LoginWithToken.tsx
setchy: Absolutely!  Saving future us heartache, thank you!

Updated to use AUTH_SCOPES array length

PR #966, opened 2024-04-06
feat(settings): show account hostname

setchy:
Allow user to specify if they want to show notification hostname.  Currently, this will only show if multiple accounts are authenticated.

Screenshot 2024-04-06 at 7 25 46 AM


afonsojramos:
I actually wanted to suggest some changes to this PR, why was there a rush to merge?

Does it make sense to have this as a setting? Why don't we just provide a "all,"X hostname","Y hostname", when there are multiple accounts? This way we don't need to add a setting that won't affect most users

PR #967, opened 2024-04-06
refactor: default to repo url

setchy:
Ensure there is always a url to open

### DIFF
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -252,7 +252,7 @@ export async function generateGitHubWebUrl(
   notification: Notification,
   accounts: AuthState,
 ): Promise<string> {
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):

src/utils/helpers.ts
```diff
@@ -252,7 +252,7 @@ export async function generateGitHubWebUrl(
   notification: Notification,
   accounts: AuthState,
 ): Promise<string> {
-  let url: string;
+  let url = notification.repository.html_url;
```
afonsojramos, line 255: Well, this is a bit of a messy situation, because the default case made more sense, but obviously it doesn't always enter that conditional... Nice bug find!

PR #968, opened 2024-04-06
feat: bring back ability to click full notification row

setchy:
Back by popular demand (and muscle memory)

### DIFF
--- a/src/components/NotificationRow.tsx
+++ b/src/components/NotificationRow.tsx
@@ -53,6 +53,13 @@ export const NotificationRow: React.FC<IProps> = ({
     unsubscribeNotification(notification.id, hostname);
   };
 
+  const openUserProfile = (event: React.MouseEvent<HTMLElement>) => {
+    // Don't trigger onClick of parent element.
+    event.stopPropagation();
+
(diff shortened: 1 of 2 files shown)


src/components/NotificationRow.tsx
```diff
@@ -80,11 +87,11 @@ export const NotificationRow: React.FC<IProps> = ({
         <NotificationIcon size={18} aria-label={notification.subject.type} />
       </div>
 
-      <div className="flex-1 overflow-hidden">
+      <div className="flex-1 overflow-hidden" onClick={() => pressTitle()}>
         <div
           className="mb-1 text-sm whitespace-nowrap overflow-ellipsis overflow-hidden cursor-pointer"
           role="main"
-          onClick={() => pressTitle()}
+          // onClick={() => pressTitle()}
```
afonsojramos, line 94:
```suggestion
```
No need to keep this, it is pretty self-explanatory if we want to rollback!

afonsojramos (approved):

src/components/NotificationRow.tsx
setchy: Ooops, left that there by accident.  Great catch

PR #1005, opened 2024-04-14
refactor: generalize color setting and skip detailed subject fetching when disabled

setchy:
Rename the existing `colors` settings to a more general `detailedNotifications`.

When set to false, skip fetching detailed subject info for each notification (help avoid rate-limiting).

Note: I have used this setting as the example within https://github.com/gitify-app/gitify/pull/1006


setchy:
A question: should this be defaulted on or off 🤔

bmulholland (approved):
Yeah this is much more clear, nice :)

afonsojramos (approved):

afonsojramos:
Let's keep it off I suppose, until we find a way to not trigger rate-limiting

PR #1007, opened 2024-04-15
fix(settings): allow showBots only when detailedNotifications is selected

setchy:
Also make it null safe for notifications that don't have users (ie: workflow runs)


afonsojramos (approved):
Seems logical I suppose!

setchy:
> Seems logical I suppose!

now with added tests :)

PR #1009, opened 2024-04-15
fix: move `update-native-theme` into ipcRenderer `useEffect`

setchy:
I've been seeing the following warning/error in the console when running locally

```
(node:7668) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 update-native-theme listeners added to [IpcRenderer]. Use emitter.setMaxListeners() to increase limit
```

Moving this into the useEffect along with the other ipcRenderer calls seems to alleviate this.  Tested by toggling theme within the Gitify app, as well as via the macOS system settings and all seemed to behave as anticipated.


afonsojramos (approved):
That makes sense

PR #1011, opened 2024-04-16
fix: relocate set request status succeeded 

setchy:
Moves the setRequestFailed(false) into the earliest location we have a successful response.

Fixes UI "flickering" when the app gets in an error state, such as rate-limiting (clicking refresh would flicker between notifications and the error page)

### DIFF
--- a/src/hooks/useNotifications.ts
+++ b/src/hooks/useNotifications.ts
@@ -96,12 +96,12 @@ export const useNotifications = (): NotificationsState => {
       }
 
       setIsFetching(true);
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):
Damn it! I hate this nasty bugs 🐛 

**SQUASHED**

bmulholland:
Nice work tracking that down!

setchy:
Thank you both.

We still have the _other_ issue where notifications will randomly disappear from the list, but come back on refresh/next interval.  😅

PR #1012, opened 2024-04-16
fix: combine fetchNotifications effect

setchy:
Reduces the number of fetchNotification calls on startup by combining into a single useEffect hook


afonsojramos (approved):
Well, that certainly wasn't helping 😅

PR #1014, opened 2024-04-16
feat(refresh): reset interval and disable manual refresh

setchy:
This PR further helps reduce the chance of rate-limiting conditions by
- resetting the automated refresh timer upon a manual refresh (default: 1 minute) to prevent manual+auto fetching being close together
- disabling the manual refresh button, if clicked, until the next internal (default: 1 minute) to prevent repeated clicks

I have also taken the opportunity to refactor the hardcoded fetch interval into our Constants file


adufr:
However, having a configurable interval could probably also help with the rate-limiting issue?
This way user would have more control over this issue, and if they report problems we could just instruct them to increase their refresh interval

setchy:
> Just sharing my opinion about this: first point looks good to me but I don't agree with the second, IMO the user should be free to refresh whenever he wants, and if he gets rate-limited because he clicked the refresh button too many times, too bad for him but that's his problem I often use this button and this change would badly affect my experience 🙁

You have a need to refresh multiple times within the minute? 🤔

setchy:
> However, having a configurable interval could probably also help with the rate-limiting issue?

Raised this as a new FR #1027 1027

setchy:
Kind of weird... `pnpm test -- --coverage --verbose` works locally, but `pnpm test -- --coverage --runInBand --verbose` isn't... 🤔

afonsojramos:
Agree with @adufr here. Refreshing manually may increase usage, but it was manual, so that's the user's responsibility. We should not take that "power" away imo.

setchy:
> Agree with @adufr here. Refreshing manually may increase usage, but it was manual, so that's the user's responsibility. We should not take that "power" away imo.

OK then....

afonsojramos:
Reverting the change in https://github.com/gitify-app/gitify/pull/1014/commits/e62aa37d0014cea3c1061ee9cb3eff0fa0218589, we can iterate in the future if needed. Let's get `5.3.0` out!

afonsojramos (approved):

PR #1030, opened 2024-04-17
feat(error): handle network connection errors

setchy:
Screenshot 2024-04-17 at 8 14 38 AM

### DIFF
--- a/src/components/Oops.test.tsx
+++ b/src/components/Oops.test.tsx
@@ -6,7 +6,7 @@ describe('components/Oops.tsx', () => {
   it('should render itself & its children', () => {
     const mockError = {
       title: 'Error title',
-      description: 'Error description',
+      descriptions: ['Error description'],
       emojis: ['🔥'],
     };
     const tree = TestRenderer.create(<Oops error={mockError} />);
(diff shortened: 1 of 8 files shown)


bmulholland (approved):

bmulholland:
Nice! I'd probably have just added a linebreak in that two-line error and translated that to a <br> or something. But maybe I'm just lazy.

setchy:
> Nice! I'd probably have just added a linebreak in that two-line error and translated that to a or something. But maybe I'm just lazy.

Fair feedback, and I did pivot 3 times to land on this 🤷 

I did briefly go down that route, but it would require something [like this](https://dev.to/kalush89/how-to-parse-html-string-in-react-53fh) - manipulating innerHTML feels smelly, and those two libraries haven't been released in _ages_

I also looked into having description support `ReactNode | string` like we do for the component/fields, but building this in the Constants file got wild with nested React.createElements

Using an array felt like the lesser of the other "evils" :p

bmulholland:
Reasoning sounds good to me :)

setchy:
@afonsojramos - thoughts? 👍 👎

afonsojramos (approved):
Reasoning seems sound and I suppose the impact isn't that bad. Let's go with it!

PR #1031, opened 2024-04-17
fix: only show `showNotificationsCountInTray` setting on MacOS

adufr:
Looks like the Electron method we're using for the `showNotificationsCountInTray` setting [is MacOS only](https://www.electronjs.org/docs/latest/api/tray#traysettitletitle-options-macos) 🙈  
image

### DIFF
--- a/src/routes/Settings.tsx
+++ b/src/routes/Settings.tsx
@@ -36,6 +36,7 @@ export const SettingsRoute: FC = () => {
   const navigate = useNavigate();
 
   const [isLinux, setIsLinux] = useState<boolean>(false);
+  const [isMacOS, setIsMacOS] = useState<boolean>(false);
   const [appVersion, setAppVersion] = useState<string | null>(null);
   const [repoScope, setRepoScope] = useState<boolean>(false);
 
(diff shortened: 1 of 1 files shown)


bmulholland (approved):

setchy:
@adufr - we should also update https://github.com/gitify-app/gitify/blob/23c1bae69ecc3a6e29ad4349928924559c9fcf21/src/context/App.tsx#L117-L122, right?

adufr:
Mmh I don't know, the setting is false by default, so this shouldn't be a problem..?

setchy:
> Mmh I don't know, the setting is false by default, so this shouldn't be a problem..?

what if someone has changed it already ;)

adufr:
> what if someone has changed it already ;) now that you're hiding the setting, they won't be able to disable it
k but how do we handle this case? should we return early in this useEffect if the platform isn't MacOS?

>  I'm also torn between hiding the setting, or disabling it and having a tooltip explaining what this is macOS only...
I don't really see the point of having a setting that you will never be able to turn on 😅

bmulholland:
Yeah, IMHO best practice here would be to hide the setting AND force it to false wherever it's used (overriding anyone who sets it to true so far, because of a future bug, etc)

Issue #995, opened 2024-04-10
[Feature request] Keyboard shortcut to open the app


adufr:
### 📝 Provide a description of the new feature

As a keyboard poweruser, I would like to be able to open/close the app with a keyboard shortcut (preferably customizable, can be disabled by default) to avoid using my trackpad/mouse.

### ➕ Additional Information

I didn't try it for now, but I think this is possible using [electron global shortcuts API](https://www.electronjs.org/docs/latest/tutorial/keyboard-shortcuts#global-shortcuts)

If you think this could be a new feature, I'd like to work on it 😁

afonsojramos:
For sure a useful one! Yeah, feel free to work on it!

PR #1032, opened 2024-04-17
feat: keyboard shortcut to open app (with setting to disable) 

adufr:
## Related issues

#995 

## Context

- Add a keyboard shortcut to show/hide the app window without having to use the mouse/trackpad- 
- Also add a setting to enable/disable this shortcut

## Discussion

- I put the shortcut string in the settings even though this PR doesn't allow the shortcut to be customized (we can add that as another new feature later)
- What do you think about the shortcut I've added by default ? (`CmdOrCtrl+Alt+G`)
- Right now the shortcut is disabled by default, do you think we could/should enable it by default?

### DIFF
--- a/src/context/App.tsx
+++ b/src/context/App.tsx
@@ -43,7 +43,7 @@ export const defaultSettings: SettingsState = {
   detailedNotifications: false,
   markAsDoneOnOpen: false,
   showAccountHostname: false,
-  kbdShortcut: 'CmdOrCtrl+Alt+G',
+  kbdShortcut: Constants.KBD_SHORTCUT_DEFAULT,
   kbdShortcutEnabled: false,
 };
 
--- a/src/context/App.tsx
+++ b/src/context/App.tsx
@@ -19,7 +19,7 @@ import {
 } from '../types';
 import { apiRequestAuth } from '../utils/api-requests';
 import { addAccount, authGitHub, getToken, getUserData } from '../utils/auth';
-import { setAutoLaunch, updateTrayTitle } from '../utils/comms';
+import { setAutoLaunch, setKbdShortcut, updateTrayTitle } from '../utils/comms';
 import Constants from '../utils/constants';
 import { generateGitHubAPIUrl } from '../utils/helpers';
 import { getNotificationCount } from '../utils/notifications';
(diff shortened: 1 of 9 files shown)


adufr:
Hello @setchy, I've just opened this PR but in DRAFT as I'm having an issue I don't know how to solve, so I'd really appreciate some help on this 😁 

Right now the feature/setting is working great, the only problem I have is, when you enable the shortcut setting, then close the app, then reopen it, the shortcut doesn't work -> you have to toggle the setting off and on for it to be registered again

I'm having trouble to make the React app that contains the state communicate with the Electron backend responsible for registering the shortcut

src/context/App.tsx
```diff
@@ -43,6 +43,8 @@ export const defaultSettings: SettingsState = {
   detailedNotifications: false,
   markAsDoneOnOpen: false,
   showAccountHostname: false,
+  kbdShortcut: 'CmdOrCtrl+Alt+G',
```
setchy, line 46: Constants may be a better home for this :)

adufr: Yup good idea, I'll create a constant that will look something like `DEFAULT_KBD_SHORTCUT` 👍

setchy:
@adufr - just ran this on my machine to get a feel for how this would work... i know this is still WIP, but sharing that on my device the shortcut briefly shows the window before it hides.  i also get some other random artifact on another display at the same time 🤔

adufr:
> @adufr - just ran this on my machine to get a feel for how this would work... i know this is still WIP, but sharing that on my device the shortcut briefly shows the window before it hides. i also get some other random artifact on another display at the same time 🤔

That's really weird 🤨 I just tried it again and it works perfectly. I even tried it with a 4 monitors setup and no issue whatsoever... 
And the code looks very clean so I really don't understand what could cause this

You're running MacOS right? Which version are you using? (I'm on 14.4.1)

setchy:
Correct, macOS 14.4.1

adufr:
Considering that a lot of things changed recently, I'll try resolving the conflicts, to see if it works, and if it doesn't I'll just close it...

setchy:
> Considering that a lot of things changed recently, I'll try resolving the conflicts, to see if it works, and if it doesn't I'll just close it...

thanks mate.  let us know if we can assist 🙏

Issue #1029, opened 2024-04-16
Recurring/duplicated system notifications 


setchy:
### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### 📝 Description

Experience recurring/duplicated system notifications on v5.3.0, in-line with each refresh interval.

Toggling the `show system notifications` setting didn't seem to take effect (notifications continued).

Disabling `show system notifications` and restarting the app seems to have helped.

### 🪜 Steps To Reproduce

Unknown 🤷‍♂️ 

### Gitify Version

5.3.0

### Operating System

macOS

### GitHub Account

GitHub.com

### 📸 Screenshots

<img width="595" alt="Screenshot 2024-04-16 at 4 09 21 PM" src="https://github.com/gitify-app/gitify/assets/386277/64b377ee-aae3-4337-9894-f5a3434d6b5a">

PR #1038, opened 2024-04-17
fix: duplicate system notifications

setchy:
Closes: #1029


afonsojramos (approved):
Hindsight 20/20... Now it makes sense...

PR #1046, opened 2024-04-18
refactor(api/errors): move error function and tests out

setchy:
Piece by piece carving useNotifications back to something bite-sized


bmulholland (approved):
Lovely extraction, nice work

PR #1056, opened 2024-04-22
refactor: extract api calls into client with types

setchy:
Refactored most uses of apiRequestAuth into a new `api/client`.  Add explicit types where missing.  Fixed semantics around Release vs ReleaseComment.  

I've named each function to be close-to that documented in https://docs.github.com/en/rest/, and included a link to the public docs where available.


src/context/App.tsx
```diff
@@ -166,11 +165,8 @@ export const AppProvider = ({ children }: { children: ReactNode }) => {
 
   const validateToken = useCallback(
     async ({ token, hostname }: AuthTokenOptions) => {
-      await apiRequestAuth(
-        `${generateGitHubAPIUrl(hostname)}notifications`,
-        'HEAD',
-        token,
-      );
+      await headNotifications(hostname, token);
```
afonsojramos, line 168: This name is weird, but correct I suppose 😅

afonsojramos (approved):
Damn, this is one high quality pr ❤️‍🔥 Love it 💥

src/context/App.tsx
setchy: Super weird.  I was trying to think _why_ we need this specific call at all...

afonsojramos: Considering that this lives under the validateToken, I assume that it does just that 😅 Testing the token with a simple and light request just to test the permissions I suppose. BUT, we could also do it the other way, and test the token directly by looking at the scopes, similar to how we were doing in the color scopes (?)

setchy: Future us problem haha

bmulholland (approved):
Agreed, this is beautiful, nice work!

setchy:
Thanks gents 🙇

PR #1063, opened 2024-04-27
refactor: replace axios.all and axios.spread

setchy:
axios.all and axios.spread are deprecated [(source)](https://github.com/axios/axios?tab=readme-ov-file#concurrency-deprecated)

Replace with Promise.all as per their example


bmulholland (approved):
Nice, almost the exact same code but a more standard approach :)

PR #1064, opened 2024-04-28
refactor: return static github api path

setchy:
🧼 Prep for a following PR


afonsojramos:
Could have been part of that said PR imho ahaha

setchy:
> Could have been part of that said PR imho ahaha 

I specifically wanted eyes on this change, and not to get lost amongst the other change 😉

I couldn't find any docs suggesting subdomain for GitHub Cloud, but wanted to be sure

PR #1070, opened 2024-04-29
refactor: simplify useNotifications

setchy:
🧼 consolidate response handling for both github cloud and github server accounts.

### DIFF
--- a/src/hooks/useNotifications.ts
+++ b/src/hooks/useNotifications.ts
@@ -105,22 +105,17 @@ export const useNotifications = (): NotificationsState => {
         ...getEnterpriseNotifications(),
       ])
         .then(([...responses]) => {
-          // Remove any undefined responses
-          const accountResponses = responses.filter(
-            (accountResponse) => !!accountResponse,
-          );
-          const accountsNotifications: AccountNotifications[] =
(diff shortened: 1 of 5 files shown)


src/hooks/useNotifications.ts
```diff
@@ -79,15 +79,13 @@ export const useNotifications = (): NotificationsState => {
...
     async (accounts: AuthState, settings: SettingsState) => {
       function getGitHubNotifications() {
-        if (!isGitHubLoggedIn(accounts)) {
-          return;
+        if (isGitHubLoggedIn(accounts)) {
+          return listNotificationsForAuthenticatedUser(
+            Constants.DEFAULT_AUTH_OPTIONS.hostname,
+            accounts.token,
+            settings,
+          );
```
afonsojramos, line 87: I'm not sure about this change. We had a explicit return before, and now it is implicit, even though we get cleaner code... What do you think?

afonsojramos (approved):

src/hooks/useNotifications.ts
setchy: Great feedback.  I've reverted this section back to it's previous explicit early return form.

PR #1080, opened 2024-05-02
refactor: broaden try/catch and add coverage

setchy:
Move the try/catch to the main entry, `getGitifySubjectDetails` and add test coverage


afonsojramos (approved):
Much better, thanks!

PR #1085, opened 2024-05-05
refactor: extract discussions api calls into client

setchy:
Extract the discussions GraphQL API call into our api/client, along with the GQL operation document and supporting helpers


afonsojramos (approved):
Nice cleanup! There's always more! ahaha 🧽

PR #1092, opened 2024-05-08
feat: delay notification removal

### LINKED ISSUE #319: Stop re-ordering notifications when I mark one as read
I find it disorienting that notifications rearrange when I am marking some of them as 'read'. Since they group by repo, if I mark one the newest notification from repo B as read, this may cause repo A to jump ahead of repo B in the list. 

This can cause mis-clicks and certainly makes me feel like the rug has been yanked out from underneath me.

I would suggest only re-ordering when new results come from the server and/or when the window is closed or blurred and subsequently opened/focused.

setchy:
Add new **Notification Settings** which allows users to "delay" notifications from being removed from Gitify upon user interactions (marked as done, marked as read, open).

When enabled, notification rows will be set temporarily as opaque until they're removal upon next refresh (scheduled or user initiated).

Defaulted to disabled.

**Demo**
https://github.com/gitify-app/gitify/assets/386277/812a2f1c-0a15-406d-91c1-00210b432387

Closes #319

### DIFF
--- a/src/utils/remove-notification.test.ts
+++ b/src/utils/remove-notification.test.ts
@@ -3,6 +3,8 @@ import {
   mockedSingleAccountNotifications,
   mockedSingleNotification,
 } from '../__mocks__/mockedData';
+import Constants from './constants';
+// import Constants from './constants';
 import { removeNotification } from './remove-notification';
 
 describe('utils/remove-notification.ts', () => {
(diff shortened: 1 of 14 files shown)


src/utils/remove-notification.test.ts
```diff
@@ -1,7 +1,10 @@
+import { mockSettings } from '../__mocks__/mock-state';
 import {
   mockedSingleAccountNotifications,
   mockedSingleNotification,
 } from '../__mocks__/mockedData';
+import Constants from './constants';
+// import Constants from './constants';
```
afonsojramos, line 7:
```suggestion
```

afonsojramos (approved):

afonsojramos:
Nice one! Interesting feature, definitely one that feels weird at first but then feels useful!

setchy:
> Nice one! Interesting feature, definitely one that feels weird at first but then feels useful!

I know right!  I think it's actually one that I might personally enable

PR #1094, opened 2024-05-08
refactor(discussions): use API to get latest comment and reply

setchy:
Instead of comparing createdAt timestamps, we can use the GraphQL query variables to fetch the latest comment and latest reply

### DIFF
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -117,10 +117,10 @@ async function getDiscussionUrl(
 
     const comments = discussion.comments.nodes;
 
-    const latestCommentId = getLatestDiscussionComment(comments)?.databaseId;
+    const latestComment = getLatestDiscussionComment(comments);
 
-    if (latestCommentId) {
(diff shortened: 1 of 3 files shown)


src/utils/helpers.ts
```diff
@@ -155,10 +156,13 @@ export function getLatestDiscussionComment(
     return null;
   }
 
-  return comments
-    .flatMap((comment) => comment.replies.nodes)
-    .concat([comments[comments.length - 1]])
-    .reduce((a, b) => (a.createdAt > b.createdAt ? a : b));
+  // Return latest reply if available
+  if (comments[0].replies.nodes.length === 1) {
+    return comments[0].replies.nodes[0];
```
afonsojramos, line 161: So it already comes in the right order? Why were we doing all of that logic then 😂

afonsojramos (approved):

src/utils/helpers.ts
setchy: Exactly!

PR #1097, opened 2024-05-08
feat: ensure app is launched minimized

### LINKED ISSUE #678: Add option to launch minimized

On Linux, I added `gitify` to my DEs autostart. Now, the window pops up during every start instead of being started in the background.

As there is no `--help` switch on the CLI, I can't check for any `--background`.

### ➕ Additional Information

_No response_

setchy:
Closes #678 

Note: I haven't been able to test this on Linux.

But what I did do was an inverse test on macOS (eg: `menubarApp.showWindow()` and confirm that the window is shown/opened on each startup).


afonsojramos:
@killermoehre can you maybe test this locally?

setchy:
> @killermoehre can you maybe test this locally?

Build artifacts: https://github.com/gitify-app/gitify/actions/runs/9007865758

killermoehre:
> @killermoehre can you maybe test this locally?

How do I test this function? I downloaded the `.AppImage`, but there is no option to toggle the behavior depending on launch via CLI and launch via Autostart.

Without any parameter, the application is actually launched in the foreground, so no, it doesn't work, yet

killermoehre:
I don't use `gitify` anymore, actually. So I'm sorry, no help here.
Also, I don't see what a video should do here. I start the application via `./Gitify-5.5.0.AppImage` and the window opens, after the update check is done.

OTOH, if I now click on the icon in the status tray, no window opens.

PR #1103, opened 2024-05-10
feat: use `last_read_at` timestamp when available

setchy:
Prefer last_read_at time when available.  Helpful in conditions where you open a notification, then mark is as unread.  Most noticeable when a decent amount of time has lapsed between reading and unreading. 

**Before**
Screenshot 2024-05-10 at 4 43 39 PM

**After**
Screenshot 2024-05-10 at 4 43 18 PM


setchy:
Hard to tell if this is a regression in the GitHub API around how they set `updated_at` timestamps... the GQL queries to fetch discussions are also failing when previously I'm sure they were OK...  Perhaps we wait a bit to see how this pans out

afonsojramos (approved):
I see, so switching between unread/read would update the notification, that makes sense!

PR #1104, opened 2024-05-10
refactor(discussions): search syntax

setchy:
🧼 further simplification - i feel we get enough filtering from the GraphQL response that we can avoid some client-side processing

- rename and move search syntax function into graphql/utils.ts
- simplify search syntax to remove updated:>timestamp+padding  
- simplify post results filtering subscription state (if you mark as done and then mark as undone, your subscription is set to IGNORED)
- fetch the first 5 discussions

### DIFF
--- a/src/utils/api/client.ts
+++ b/src/utils/api/client.ts
@@ -20,7 +20,8 @@ import { apiRequestAuth } from './request';
 import { print } from 'graphql/language/printer';
 import Constants from '../constants';
 import { QUERY_SEARCH_DISCUSSIONS } from './graphql/discussions';
-import { formatSearchQueryString, getGitHubAPIBaseUrl } from './utils';
+import { formatAsGitHubSearchSyntax } from './graphql/utils';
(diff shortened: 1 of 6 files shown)


src/utils/api/client.ts
```diff
@@ -249,14 +250,30 @@ export async function searchDiscussions(
...
     query: print(QUERY_SEARCH_DISCUSSIONS),
     variables: {
-      queryStatement: formatSearchQueryString(
+      queryStatement: formatAsGitHubSearchSyntax(
         notification.repository.full_name,
         notification.subject.title,
-        notification.updated_at,
       ),
-      firstDiscussions: 10,
+      firstDiscussions: 1,
```
afonsojramos, line 257: What's the impact of this?

afonsojramos: Nevermind, got it, we were never looking at the rest of the comments, only to the latest

afonsojramos (approved):

PR #1112, opened 2024-05-14
feat: improve enterprise login

mashiro:
Fixed the following issues with GitHub Enterprise.

1. Settings now support logging in with pat, or logging in with oauth app, after the initial login.
2. Fixed a bug that caused the settings page to go blank when logged in only to github enterprise. Closes https://github.com/gitify-app/gitify/issues/1130
3. ~Multiple logins to github enterprise are now possible with a personal token and the github app.~ planned to 5.7.x


src/routes/LoginEnterprise.tsx
```diff
@@ -52,19 +50,9 @@ export const validate = (values: IValues): IFormErrors => {
...
 export const LoginEnterpriseRoute: FC = () => {
-  const {
-    accounts: { enterpriseAccounts },
-    loginEnterprise,
-  } = useContext(AppContext);
+  const { loginEnterprise } = useContext(AppContext);
   const navigate = useNavigate();
 
-  useEffect(() => {
-    if (enterpriseAccounts.length) {
```
setchy, line 62: I love that we're extending this to support multiple enterprise accounts. Awesome!

This logic which limits Enterprise Accounts to at most 1 is here since there currently isn't a way to view all accounts, or edit/remove accounts.  

I have been dabbling on two similar changes

- View accounts: https://github.com/gitify-app/gitify/tree/feature/accounts-view
- Support multiple accounts (GitHub or GitHub Enterprise) - https://github.com/gitify-app/gitify/tree/refactor/accounts

My suggestion for this PR would be to defer this change until the above items are addressed

setchy (changes requested):

main.js
```diff
@@ -41,6 +41,8 @@ const mb = menubar({
 });
 
 mb.on('ready', () => {
+  require('@electron/remote/main').enable(mb.window.webContents);
```
setchy, line 44: Correct me if I'm wrong, but this looks to re-introduce electron/remote, which we're wanting to remove in https://github.com/gitify-app/gitify/issues/650

@brandonweiss also has a draft PR for fixing our auth windows - https://github.com/gitify-app/gitify/pull/654

mashiro: I'm not well-versed in this implementation, and as long as authentication via PAT is possible, it's sufficient for me. Therefore, it's fine to exclude this part from the implementation.

src/routes/LoginEnterprise.test.tsx
```diff
@@ -75,35 +74,6 @@ describe('routes/LoginEnterprise.tsx', () => {
     expect(validate(values).clientSecret).toBe('Invalid client secret.');
   });
 
-  it('should receive a logged-in enterprise account', () => {
```
setchy, line 78: is there a reason why this test case has been removed?

src/routes/Settings.tsx
```diff
@@ -1,5 +1,6 @@
 import {
```
setchy, line 1: I like the changes you've made to Settings @mashiro

src/routes/Settings.tsx
```diff
@@ -75,13 +77,15 @@ export const SettingsRoute: FC = () => {
 
   useMemo(() => {
     (async () => {
-      const response = await getRootHypermediaLinks(
-        Constants.DEFAULT_AUTH_OPTIONS.hostname,
```
setchy, line 79: perhaps this should get the RootHypermediaLinks by conditionally pass the hostname (eg: GitHub ?? EnterpriseAccounts)

src/routes/LoginEnterprise.test.tsx
mashiro: It was a test for page transition when already logged in with enterprise.
Since I removed that restriction, this test case is no longer needed.

src/routes/Settings.tsx
mashiro: I considered addressing this, but it would introduce the concept of a primary account, which could complicate the implementation. Therefore, I opted for the simplest solution.

setchy: @afonsojramos - thoughts on whether the scopes check is still required?  A few months back I did do a lot of testing locally and couldn't find a scenario where I didn't have permissions to make the necessary API calls for detailed notifications (previously called colors).  Removing this would simplify the above convo

afonsojramos: Sorry for the late reply, had a couple of intense days

setchy: I agree

I think we're pretty well covered with:
* dedicated error page for Missing Scopes
* Settings tooltips explaining whats required
* new login flow via `Generate a PAT` selects all the required scopes

I'll raise a PR shortly to simplify this

setchy: https://github.com/gitify-app/gitify/pull/1127

setchy:
thank for your patience @mashiro.  fyi - i'll push an update to this feature branch shortly to help move this enhancement along.  watch this space 😃

setchy:
Updates pushed.  @mashiro @afonsojramos - appreciate some feedback.  

the Settings changes now support logging in with pat, or logging in with oauth app, after the initial login.

it is still constrained to at-most one of each.  in v5.7.x we'll focus on adding account views and a new account type model to support adding any number of each. 💪

mashiro:
@setchy Looks good!

setchy (approved):
Thank you @mashiro ❤️

PR #1142, opened 2024-05-25
refactor(mocks): use consistent naming

setchy:
Refactor mocks to use consistent naming convention. 

I also moved github response mocks under the `/api/` folder so they don't get intermingled with application mocks


afonsojramos (changes requested):
I don't think naming things `mocked` over `mock` is the standard... Take a look at vitest's mocking section. They are mocks, not a "mocked" X. Same goes for Jest, mockFn, not mockedFn 🤔

setchy:
> I don't think naming things `mocked` over `mock` is the standard... Take a look at vitest's mocking section. They are mocks, not a "mocked" X. Same goes for Jest, mockFn, not mockedFn 🤔

Take 2 😄

afonsojramos (approved):

Issue #1166, opened 2024-05-31
Gitify blank/white screen with loading bar on top than stop


renanfranca:
### 🔍 Is there already an issue for your problem?

- [X] I have checked older issues, open and closed

### 📝 Description

I can't see the notifications list because the screen is all white. I tried uninstall and install and did not solve the problem.

### 🪜 Steps To Reproduce

1 - Go to the windows trails icon
2 - Click on Gitify trail Icon
3 - You will see a blank/white window
4 - I cant go to settings or do anything

### Gitify Version

5.6.0

### Operating System

Windows

### GitHub Account

GitHub.com

### 📸 Screenshots

![image](https://github.com/gitify-app/gitify/assets/6011616/1d0c9c0e-1c99-464a-ad2a-7dcdbfb38965)

pddg:
Same here on macOS. The version is macOS Sonoma 14.5.
<img width="523" alt="スクリーンショット 2024-06-01 8 32 58" src="https://github.com/gitify-app/gitify/assets/14105201/5851fd22-974d-42d6-bcd5-4e002367c9f7">

Notification of how many notifications there are came at startup. However, it stays on this screen after startup and cannot be terminated.
I have tried reinstalling, etc., but nothing has changed.

setchy:
Thanks for reporting. 

To help triage further, could you both share the console logs from debug mode - https://www.gitify.io/faq/

renanfranca:
You are welcome!

@setchy : The shortcut to debug mode on windows is `ctrl + shift + i`. Here is the print screen from console and network:

![image](https://github.com/gitify-app/gitify/assets/6011616/0567e041-2df6-48e6-9b92-517ddc60fa06)

setchy:
> @setchy : The shortcut to debug mode on windows is `ctrl + shift + i`.

Thank you 🙏. FAQ has been updated

setchy:
fix raised: https://github.com/gitify-app/gitify/pull/1167

renanfranca:
> Thank you 🙏. FAQ has been updated

I'm glad to help! 😃

PR #1167, opened 2024-06-01
fix: add datetime error handling

setchy:
I'm take a bit of an educated guess, but this should catch any null or split issues with our use of date-fns lib

Fixes: #1166


afonsojramos (approved):
That seemed to fix it indeed!

xShteff:
Would it perhaps make sense to release 5.6.1 patch with this fix, to fix people's installs? (:

setchy:
> Would it perhaps make sense to release 5.6.1 patch with this fix, to fix people's installs? (:

We'll be releasing 5.7.0 early this week :)

PR #1172, opened 2024-06-03
feat: new account management view

setchy:


afonsojramos:
Looks really cool! Awesome actually! One thing I'd change is to only show the icon in the sidebar in case there are multiple accounts already, what do you think?

setchy:
We could move the Account Management button into the Settings footer, in place of where the "add extra accounts" buttons currently are... It would leave the sidebar as-is (win), but, does add another layer of settings mgmt depth 🤷‍♂️

setchy:
> One possible suggestion...we _could_ decide to move the Account Management button into the Settings footer, in place of where the "add extra accounts" buttons currently are... It would leave the sidebar as-is (win), but, does add another layer of settings mgmt depth 🤷‍♂️

decided to go down this route... branch updated :)

afonsojramos:
@setchy Looking good! How about we remove the "master" Logout? 🧐

setchy:
> @setchy Looking good! How about we remove the "master" Logout? 🧐

Good callout.  Done 🧹 🧼

bmulholland (approved):
Code is good to my eyes. Really nice work in all the steps getting here, too. Great stuff!

>  Coverage decreased (-0.07%) to 96.904% 

Demanding!

setchy:
> >  Coverage decreased (-0.07%) to 96.904% 
> 
> So demanding! Not an issue for me.

Coveralls really has no chill haha. It dropped when removing the master logout. 

96+% coverage is still a mighty stat to be proud of

afonsojramos (approved):
Looking good! Excited to see this live!

PR #1173, opened 2024-06-04
refactor: user's name can be null

setchy:
improve typing for user


afonsojramos (approved):
Had to double check that, but checks out!

setchy:
> Had to double check that, but checks out!

Haha thx, yeah, found this today with my test github account.

PR #1177, opened 2024-06-04
refactor: update notification footer styles

setchy:
handful of small tailwind tweaks

- prevent avatar from being shrunk on overflow
- allow horizontal scrolling on overflow (for notifications with lots of pills. This is a little quirky, so open to suggestions on a better method)
- simplify margins by using gap

### DIFF
--- a/src/components/NotificationRow.tsx
+++ b/src/components/NotificationRow.tsx
@@ -114,111 +114,104 @@ export const NotificationRow: FC<IProps> = ({ notification, hostname }) => {
       </div>
 
       <div
-        className="flex-1 overflow-hidden"
+        className="flex-1 whitespace-nowrap overflow-hidden overflow-ellipsis"
         onClick={() => openNotification()}
         onKeyDown={() => openNotification()}
       >
         <div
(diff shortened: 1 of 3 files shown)


afonsojramos (approved):

afonsojramos:
Before merging, do you have an example of an overflow? Might be worth looking over

setchy:
> Before merging, do you have an example of an overflow? Might be worth looking over

https://github.com/gitify-app/gitify/assets/386277/ac96bf70-c6de-4be8-bae6-3f88dbb199b2

setchy:
would be nice to use less space for the update type and formatted datetime... but couldn't think of anything clean yet...

bmulholland:
I don't think I'd ever notice that horizontal scroll myself. Wrapping to new line isn't an option?

bmulholland (approved):
Regardless of UX, this code is a strict improvement and is good to go, IMO. Other changes discussed could/should be a separate step, I think.

setchy:
> Regardless of UX, I believe this code is a strict improvement and is good to go. IMHO.,other changes discussed above could/should be a separate step.

Thanks @bmulholland. I think this is pragmatic and affords us more time to find alternative solutions without holding up the next release

PR #1186, opened 2024-06-05
fix: tooltip z index to prevent items showing through

setchy:
**Before**
Screenshot 2024-06-05 at 12 53 38 PM

**After**
Screenshot 2024-06-05 at 12 53 09 PM


bmulholland:
LGTM, nice fix

PR #1201, opened 2024-06-08
feat: support multiple accounts

### LINKED ISSUE #365: Login to multiple Github accounts simultaneously
Hi, first of all, thank you for this awesome application!

I have a feature request: I am using Github personally and during work, where I have 2 Github handles. It would be great to be able to login to multiple accounts and receive notifications from all of them.

setchy:
This PR adds support for multiple GitHub Cloud or GitHub Enterprise Service accounts. 🎉 

Majority of the code changes were to:
- pass notification object instead of its parts 
- use account.token instead of deriving it from hostname (now that hostname isn't unique)
- use unique `getAccountUUID` for notification interactions (mark as read, raise native notifications, etc) so that state is correctly managed per account and not per hostname

Closes #365


setchy:
I think this is ready for review and ideally further local testing.

setchy:
looking into the test coverage drop 👀

afonsojramos (approved):

afonsojramos:
Locally tested and everything looks awesome! One edge case that we can improve is:

image

When using the same account twice (same PAT even), the dropdown doesn't work, although it does show.

image

setchy:
> One edge case that we can improve

ha - you found the edge case straight away...  yep, haven't implemented any duplicate checking - figured that can just be handled by end-users :)

Issue #650, opened 2023-10-03
Remove @electron/remote


bmulholland:
[Usage of @electron/remote is discouraged](https://github.com/electron/remote/blob/main/README.md#electronremote)

We can replace it with inter-process communication fairly easily, just need to do it piece by piece.

setchy:
@bmulholland - would you mind confirming that this can be closed.  I did a search and couldn't find any references to `remote` from the `electron` package directly.  All seem to have been migrated to `@electron/remote`

bmulholland:
> All seem to have been migrated to `@electron/remote`

That's the package that is discouraged and shouldn't be used. From the linked article, about `@electron/remote` itself:

> ⚠️ Warning! This module has [many subtle pitfalls](https://medium.com/@nornagon/electrons-remote-module-considered-harmful-70d69500f31). There is almost always a better way to accomplish your task than using this module. For example, [ipcRenderer.invoke](https://www.electronjs.org/docs/latest/api/ipc-renderer#ipcrendererinvokechannel-args) can serve many common use cases.

setchy:
Oops!  I misread it.  I focused on this comment but missed the warning.  Thanks for the reminder
> @electron/remote is a replacement for the built-in remote module in Electron, which is deprecated and will eventually be removed.

setchy:
See https://www.electronjs.org/docs/latest/api/context-bridge

setchy:
@electron/remote no longer used https://github.com/gitify-app/gitify/pull/1781

PR #1207, opened 2024-06-08
refactor: main js

setchy:
update to closer match structure of [examples](https://www.electronjs.org/docs/latest/tutorial/examples)


main.js
```diff
@@ -1,18 +1,17 @@
-const { ipcMain: ipc, app, nativeTheme } = require('electron');
+const { ipcMain: ipc, app, nativeTheme } = require('electron/main');
 const { menubar } = require('menubar');
 const { autoUpdater } = require('electron-updater');
 const { onFirstRunMaybe } = require('./first-run');
 const path = require('node:path');
 
-require('@electron/remote/main').initialize();
```
setchy, line 7: i dont think this is required...

bmulholland (approved):
Huge fan of this change. These types of updates are easy to overlook, but often solve hours of unnecessary debugging. Thanks!

main.js
setchy, line 7: @bmulholland - thoughts on if this is OK to remove now.  I know you were cleaning this up via https://github.com/gitify-app/gitify/pull/654

afonsojramos: If it doesn't work it doesn't, but otherwise I see no reason to not remove it!

setchy: I don't want to pull the trigger on this yet.  left a TODO referencing #650 for future us

PR #1212, opened 2024-06-09
feat: tray context menu

setchy:
Add right-click context menu with `Quit` app option


afonsojramos (approved):
It is not that common that menubar items have right-click specific menus, BUT, there are some indeed. So I think that we can move forward with this! We can find more stuff to put this menu in the future

bmulholland:
Refresh might be a fit for another item?

PR #1213, opened 2024-06-09
build(release): remove gh_token env

setchy:
~use default auto provisioned GitHub actions token instead of my custom PAT~

remove gh_token env variable. should use `github_token`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
           APPLEID_TEAM_ID: ${{ secrets.appleid_teamid }}
           CSC_LINK: ${{ secrets.mac_certs }}
           CSC_KEY_PASSWORD: ${{ secrets.mac_certs_password }}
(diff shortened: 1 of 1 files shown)


afonsojramos (changes requested):
The default default GitHub actions token, afaik doesn't need to be provided at all, so we can just delete it I think!

setchy:
> The default default GitHub actions token, afaik doesn't need to be provided at all, so we can just delete it I think!

even better.  i was trying to recall when this got added.  i think it was during our notorization dark days haha

setchy:
looking again, i think it's needed so that it's passed as process.env...

setchy:
no dice - had to revert this for the 5.8.0 release... without specifying GITHUB_TOKEN or GH_TOKEN, electron-builder doesn't work.  using GITHUB_TOKEN doesn't have enough permissions to upload artifacts to the release notes.  Will continue using GH_TOKEN for now.  Ideally we'd create this within the gitify-app org

PR #1214, opened 2024-06-10
refactor: move custom styles into class

setchy:
Move style constants into dedicated class.


afonsojramos (approved):

bmulholland:
FWIW, my experience with Tailwind is that these are usually centralized by turning the full tag into a component, centralizing the HTML along with the classes.

setchy:
@bmulholland - agree...  ideally we'd have these as components.  feel free to capture that as a gh issue.

PR #1223, opened 2024-06-11
refactor: introduce branded types

dammy95:
Closes https://github.com/gitify-app/gitify/issues/1141

Created the following branded types:

- [X]  `Hostname`
- [X] `Token`
- [X] `ClientSecret`
- [X] `ClientID`
- [X] `AuthCode`
- [x] `URL` (see follow-up question below)

Follow-up question:
While attempting to create the `URL` branded type, I observed it clashes with JavaScript's URL class. There are also different variations of `url` (i.e `thread_url`, `html_url`, `upload_url`, etc) in the project that I'm unsure of how to handle – do I ignore these variations and create the branded type for the occurrences with `url: string`?


src/utils/branded-types.ts
```diff
@@ -0,0 +1,15 @@
+declare const __brand: unique symbol;
```
setchy, line 1: Nice!  

Could we move these into `src/types.ts`

src/utils/branded-types.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type AuthCode = Branded<string, 'AuthCode'>;
+
+export type Token = Branded<string, 'Token'>;
+
+export type ClientID = Branded<string, 'ClientID'>;
+
+export type ClientSecret = Branded<string, 'ClientSecret'>;
+
+export type HostName = Branded<string, 'HostName'>;
```
setchy, line 15: Could we please name this `Hostname` vs `HostName`

setchy (changes requested):
Awesome @dammy95 - left a few comments

setchy:
Awesome work @dammy95 - thank you 🙏 

re: `Url` - all of those are full html urls...  Maybe to avoid the conflict with the JavaScript URL class we can name it differently... perhaps `Link` or `WebUrl`.  @afonsojramos @bmulholland - what do you guys think?

bmulholland:
Link reads less awkwardly than WebUrl, at least to me. I have no better ideas.

afonsojramos:
Yep, agree with the above!

dammy95:
Thanks for the feedback @setchy @afonsojramos @bmulholland ⭐

src/utils/branded-types.ts
dammy95: All done 🙏🏾

src/utils/branded-types.ts
dammy95: Done 🙏🏾

setchy (approved):
Thanks @dammy95

PR #1226, opened 2024-06-12
refactor: biome linter updates

setchy:
A selection of updates to remove biome warnings, including:
- ensure all callbacks specify their dependencies
- enable `noUnusedFunctionParamaters` nursery rule and fix lint errors
- enable `useDefaultSwitchClause` nursery rule and fix lint errors

### DIFF
--- a/src/components/NotificationRow.tsx
+++ b/src/components/NotificationRow.tsx
@@ -11,7 +11,7 @@ import {
 import { type FC, type MouseEvent, useCallback, useContext } from 'react';
 
 import { AppContext } from '../context/App';
-import { type Account, IconColor } from '../types';
+import { IconColor } from '../types';
 import type { Notification } from '../typesGitHub';
 import {
   formatForDisplay,
@@ -27,19 +27,16 @@ import { formatReason } from '../utils/reason';
 import { PillButton } from './buttons/PillButton';
 
 interface IProps {
-  account: Account;
   notification: Notification;
 }
 
-export const NotificationRow: FC<IProps> = ({ notification, account }) => {
+export const NotificationRow: FC<IProps> = ({ notification }) => {
   const {
-    auth,
     settings,
     removeNotificationFromState,
     markNotificationRead,
     markNotificationDone,
     unsubscribeNotification,
-    notifications,
   } = useContext(AppContext);
 
   const handleNotification = useCallback(() => {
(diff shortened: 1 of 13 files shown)


src/components/NotificationRow.tsx
```diff
@@ -51,8 +48,12 @@ export const NotificationRow: FC<IProps> = ({ notification, account }) => {
       // no need to mark as read, github does it by default when opening it
       removeNotificationFromState(settings, notification);
     }
-  }, [notifications, notification, auth, settings]); // notifications required here to prevent weird state issues
-
+  }, [
+    notification, // notifications required here to prevent weird state issues
```
setchy, line 52: Can this comment be removed??

src/components/NotificationRow.tsx
```diff
@@ -59,8 +56,12 @@ export const NotificationRow: FC<IProps> = ({ notification, account }) => {
       // no need to mark as read, github does it by default when opening it
       removeNotificationFromState(settings, notification);
     }
-  }, [notifications, notification, auth, settings]); // notifications required here to prevent weird state issues
```
bmulholland, line 62: You're sure the state issues won't come back? :)

bmulholland (approved):
Seems more explicit now, which is good. Just some risk in introducing state issues :)

src/components/NotificationRow.tsx
setchy: That comment I left "future us" has been very much front of mind...

I think adding `notifications` was a hack because the true dependencies weren't being specified.

Been running this branch locally and so far it's been behaving as expected (touch wood)

setchy:
@afonsojramos - appreciate your feedback on this too when you get a chance 🙏

afonsojramos (approved):
As most are dependency arrays for callbacks I think it should be fine!

afonsojramos:
According to React docs:

> dependencies: The list of all reactive values referenced inside of the fn code. Reactive values include props, state, and all the variables and functions declared directly inside your component body. If your linter is configured for React, it will verify that every reactive value is correctly specified as a dependency. The list of dependencies must have a constant number of items and be written inline like [dep1, dep2, dep3]. React will compare each dependency with its previous value using the Object.is comparison algorithm.

https://react.dev/reference/react/useCallback

Seems to me that passing these functions as dependencies won't have any impact tbh...

Issue #980, opened 2024-04-07
Developer Settings


setchy:
### 📝 Provide a description of the new feature

Been thinking about adding a new `Settings` group called `Developer Settings`, which could help maintainers and/or power users with debugging any issues

Settings could include
- Take app screenshot (leverage https://github.com/gitify-app/gitify/tree/feat/website-screenshot-helper)
- Generate debug.json (save GitifyNotifications object - dependent on  #828)
- Reset app (remove FirstRun folder, clear any saved credentials/settings, etc)

PR #1237, opened 2024-06-14
feat: set app version as `dev` when building locally

setchy:
A small enhancement to help when doing local development remember which version of the app is running 😎 

Screenshot 2024-06-14 at 9 39 24 AM

### DIFF
--- a/src/routes/Settings.test.tsx
+++ b/src/routes/Settings.test.tsx
@@ -31,6 +31,11 @@ describe('routes/Settings.tsx', () => {
     Object.defineProperty(process, 'platform', {
       value: originalPlatform,
     });
+
+    // Restore the original node env value
+    Object.defineProperty(process, 'platform', {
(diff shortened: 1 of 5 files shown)


bmulholland (approved):
Good idea. Could also consider changing a key color, like the sidebar background, so it's hard to miss

afonsojramos:
Would also be nice to have "demo notifications" so that we don't need to load real ones 🤔

setchy:
I have a `stash` locally that mocks the API response.  Could add that into the scope of #980

setchy:
A while ago I was using https://github.com/setchy/gitify-notifications-sandbox too, which could be transferred into the `gitify-app` organization if we wanted "real" notifications

afonsojramos:
I suppose there's the two cases, the ones where we just want to work with UI, and the cases where we need to work with the flow themselves.

PR #1232, opened 2024-06-13
refactor: deprecate react-transition-group over tailwind animations

afonsojramos:
Resolves #1071.

PR #1241, opened 2024-06-14
fix: regression in delay notification state behavior

setchy:
fix regression issue from #1232 when using `delayNotificationState` setting. 

**_Before_** 

https://github.com/gitify-app/gitify/assets/386277/f90effda-9da2-485f-bf50-f617ff46a739

**_After_** 

https://github.com/gitify-app/gitify/assets/386277/02096ffa-9bfa-4924-a027-5ee73993f77c

### DIFF
--- a/src/utils/remove-notification.ts
+++ b/src/utils/remove-notification.ts
@@ -1,4 +1,3 @@
-import { READ_NOTIFICATION_CLASS_NAME } from '../styles/gitify';
 import type { AccountNotifications, SettingsState } from '../types';
 import type { Notification } from '../typesGitHub';
 import { getAccountUUID } from './auth/utils';
@@ -9,8 +8,6 @@ export const removeNotification = (
   notifications: AccountNotifications[],
 ): AccountNotifications[] => {
   if (settings.delayNotificationState) {
-    const notificationRow = document.getElementById(notification.id);
-    notificationRow.className += ` ${READ_NOTIFICATION_CLASS_NAME}`;
     return notifications;
   }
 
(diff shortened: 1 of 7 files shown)


afonsojramos:
What is the regression exactly? Does this not remove the notification before the animation ends?

setchy:
The regression is the row doesn't get set as opaque and left until the next refresh interval.

Instead, it's removed but left with a blank gap/void. Particularly noticable when you interact with the last notification as it leaves the repository heading with a void

setchy:
You'll see pretty quickly if you enable that flag 🙂

setchy:
Happy to hear if there is a better way 😎

setchy:
https://github.com/gitify-app/gitify/assets/386277/6e72d8af-a093-484f-80ef-8d72e7323647

src/utils/remove-notification.ts
```diff
@@ -14,6 +17,9 @@ export const removeNotification = (
     return notifications;
   }
 
+  const notificationRow = document.getElementById(notification.id);
+  notificationRow.className += ` ${ANIMATE_NOTIFICATION_CLASS_NAME}`;
```
afonsojramos, line 21: The thing is that this breaks the animation though. If you use gitify without this setting the animation is now instant and you don't see it leaving. The animation needs the animation time before removing the notification.

I suppose what we can do is add a delay in this file if `delayNotificationState` is not enabled. We already have access to the setting state!

afonsojramos (changes requested):

src/utils/remove-notification.ts
setchy: Good point.

Update pushed.  I did notice that the animations are pretty quick ⚡

afonsojramos: Checking it out atm, but like the approach code wise!

setchy: Any further changes or good to roll?

afonsojramos: Sorry, yes, good to go! I've had a few visual anomalies, but I don't think they're related, it is just webpack or something from running locally.

afonsojramos (approved):

PR #1238, opened 2024-06-14
style: cleanup useless tw classnames

afonsojramos:
Cleanup a bunch of useless `flex-1`s and centralize the global `bg` and `text` colors into `index.html`.

Also fixed a bug in the login pages where the label was not being correctly set.
image

PR #1242, opened 2024-06-14
fix: regression in all read ui

setchy:
fix regression from #1238 for AllRead UI not being vertically centered


afonsojramos (approved):

afonsojramos:
Whoops

PR #1245, opened 2024-06-15
feat: toggle repository notification

### LINKED ISSUE #1243: Chevron on repository header to collapse all notifications

A way to collapse (hide) all notifications from a specific repository, while keeping the repository header

maybe something like this:
image

### ➕ Additional Information

_No response_

setchy:
Closes #1243 

Screenshot 2024-06-15 at 12 30 32 PM

### DIFF
--- a/src/components/AccountNotifications.tsx
+++ b/src/components/AccountNotifications.tsx
@@ -31,30 +31,31 @@ export const AccountNotifications = (props: IProps) => {
     ),
   );
 
-  const [showNotifications, setShowNotifications] = useState(true);
+  const [showAccountNotifications, setShowAccountNotifications] =
+    useState(true);
 
-  const toggleNotifications = () => {
-    setShowNotifications(!showNotifications);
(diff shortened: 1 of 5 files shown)


afonsojramos (commented):
Works well, but I think the UX is a bit weird since the Chevron location is not consistent (account vs repository).

Do you think that putting it on the left would look weird? Also, I think that this should be the default behaviour for clicking the row itself, what do you think?

setchy:
> Do you think that putting it on the left would look weird? 

that would be the ideal spot.  however, it then means that the `mark as read` and `mark as done` buttons are no longer aligned between repository and notification rows.  having it on the left feels like the lesser of the two evils :) 

> Also, I think that this should be the default behaviour for clicking the row itself, what do you think?

i like this suggestion - perhaps in a separate PR and applied to the Account and Repository rows

afonsojramos (approved):

PR #1246, opened 2024-06-15
build: move lint into separate workflow 

setchy:


afonsojramos (changes requested):
Hmmm, why exactly?

setchy:
> Hmmm, why exactly?

Explicit check feedback

afonsojramos:
> > Hmmm, why exactly?
> 
> Explicit check feedback 

Isn't looking at the workflow failure reason not enough? 😅

setchy:
> > > Hmmm, why exactly?
> > 
> > Explicit check feedback 
> 
> Isn't looking at the workflow failure reason not enough? 😅

I think it's more intuitive if it is a separate step, whether that's a separate workflow file (lint.yml) or the same (test.yml).

We have this clarity already in Build App and Release App and Validators 😎

afonsojramos (approved):
Why not :D

bmulholland:
IMHO this is best practice :)

PR #1247, opened 2024-06-15
build: use macos-latest runners

setchy:
Update to use `macos-latest` tag now that it's equal to `macos-14` now

Also update our tests workflow to run on macos instead of ubuntu, since the maintainer group are all macos users.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   run-unit-tests:
     name: Run Tests
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   run-unit-tests:
     name: Run Tests
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
afonsojramos, line 12: We're lucky we're open source, because the MacOS multiplier for actions minutes is crazy 😬

Outside of this, I think it is good practice to only **not** use Linux if really required.

afonsojramos (approved):

PR #1248, opened 2024-06-16
style: toggle on row click & action buttons order

afonsojramos:
Adds the functionality that was referred to in https://github.com/gitify-app/gitify/pull/1245#issuecomment-2171428582

Also sorted the order of the buttons to try to attain some added consistency.

image

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -11,6 +11,9 @@
         "noUnusedFunctionParameters": "error",
         "useDefaultSwitchClause": "error"
       },
+      "a11y": {
+        "useKeyWithClickEvents": "off"
+      },
       "correctness": {
         "useExhaustiveDependencies": {
           "level": "warn",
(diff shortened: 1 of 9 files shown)


afonsojramos:
@setchy what do you think of this one?

setchy (approved):

setchy:
> @setchy what do you think of this one?

clever ordering hack.  does mean we diverge a bit from GitHub UI, but worth it imo

afonsojramos:
How so? Because here the disable notification is right next to the check? Don't think that's deviating too much tbh
image

setchy:
> How so? Because here the disable notification is right next to the check? Don't think that's deviating too much tbh
> image
> 

Just recalling a previous PR discussion on these actions buttons. 

I'm good with it

PR #1251, opened 2024-06-16
style: notification type icon size

afonsojramos:
Notification type icon was a bit too big from the information architecture perspective, so I reduced it to the size of the rest of the icons.

Before
image

After
image

Note: After includes fixes from other PRs.

### DIFF
--- a/src/components/NotificationRow.tsx
+++ b/src/components/NotificationRow.tsx
@@ -108,7 +108,7 @@ export const NotificationRow: FC<IProps> = ({ notification }) => {
         className={cn('mr-3 flex w-5 items-center justify-center', iconColor)}
         title={notificationTitle}
       >
-        <NotificationIcon size={18} aria-label={notification.subject.type} />
+        <NotificationIcon size={14} aria-label={notification.subject.type} />
       </div>
 
       <div
@@ -154,7 +154,7 @@ export const NotificationRow: FC<IProps> = ({ notification }) => {
           <div title={reason.description}>{reason.title}</div>
           <div title={updatedLabel}>{updatedAt}</div>
           {settings.showPills && (
-            <div>
+            <div className="flex">
               {notification.subject?.linkedIssues?.length > 0 && (
                 <PillButton
                   title={linkedIssuesPillDescription}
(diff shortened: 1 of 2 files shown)


setchy:
I kinda liked that the notification type icon was larger 🤷‍♂️

src/components/NotificationRow.tsx
```diff
@@ -108,7 +108,7 @@ export const NotificationRow: FC<IProps> = ({ notification }) => {
         className={cn('mr-3 flex w-5 items-center justify-center', iconColor)}
         title={notificationTitle}
       >
-        <NotificationIcon size={18} aria-label={notification.subject.type} />
+        <NotificationIcon size={14} aria-label={notification.subject.type} />
```
afonsojramos, line 111:
```suggestion
        <NotificationIcon size={16} aria-label={notification.subject.type} />
```
@setchy let's agree on this middleground?

src/components/NotificationRow.tsx
```diff
@@ -154,7 +154,7 @@ export const NotificationRow: FC<IProps> = ({ notification }) => {
           <div title={reason.description}>{reason.title}</div>
           <div title={updatedLabel}>{updatedAt}</div>
           {settings.showPills && (
-            <div>
+            <div className="flex">
```
afonsojramos, line 157: @setchy this actually fixes the pill regression! Was supposed to be on the other PR :(

setchy: Ah, got it. I'll let you raise the fix with this and I'll close mine

setchy: Though this does mean that the PillButton isn't usable standalone and requires the consuming component to apply that...

afonsojramos: I understand what you are saying, but the use of gap is always superior to the use of margins.

setchy (approved):

afonsojramos:
> Thanks for thinking about -- and acting on! -- the information hierarchy!

I live with a designer 👀, so I always try to squeeze more opinions about Gitify out of her! 🎨

PR #1260, opened 2024-06-17
feat(settings): keyboard shortcut to open app

setchy:
revive #1032

closes https://github.com/gitify-app/gitify/issues/995

Changed the shortcut key to be `CommandOrControl+Shift+G` since `CmdOrCtrl+Alt+G` is taken by Google Drive on macOS (this explains the previous flickering I reported on #1032).

Screenshot 2024-06-17 at 6 07 12 PM


afonsojramos:
What if we give the ability to set the shortcut they want?

setchy:
> What if we give the ability to set the shortcut they want?

Could that be a separate enhancement?

main.js
```diff
@@ -122,6 +128,27 @@ app.whenReady().then(async () => {
     }
   });
 
+  ipc.on(
+    'gitify:update-keyboard-shortcut',
+    (_, { enabled, keyboardShortcut }) => {
+      console.log('enabled', enabled);
+
+      console.log('keyboardShortcut', keyboardShortcut);
```
bmulholland, line 136: Should these still be here? :)

bmulholland (approved):

main.js
setchy: not at all 🙈

setchy: Removed 🧼 

Great catch @bmulholland  🙇

afonsojramos (approved):

PR #1264, opened 2024-06-18
refactor: functions over arrow fns

setchy:
convert several arrow fns to regular fns to code consistency


bmulholland (approved):
Worth a biome rule or something to enforce this? Otherwise it'll drift again over time :)

PR #1268, opened 2024-06-18
build(biome): enable noConsoleLog rule

setchy:
Avoid future `console.logs` accidentally creeping in 🙈


scripts/notarize.js
```diff
@@ -10,12 +10,14 @@ const notarizeApp = async (context) => {
   const shouldNotarize = process.env.NOTARIZE === 'true';
 
   if (!shouldNotarize || !isMacOS) {
+    // biome-ignore lint/suspicious/noConsoleLog: console.log is used for debugging
```
bmulholland, line 13: Idea: formalize this into a "debugLog" function, silence biome once, gain better clarity on intent.

bmulholland (approved):

scripts/notarize.js
setchy: Done 🤝

afonsojramos (approved):
The fact that you refactored a console log into its own function to not put the ignore on multiple places is a bit hilarious 😆

setchy:
> The fact that you refactored a console log into its own function to not put the ignore on multiple places is a bit hilarious 😆

just making our friend @bmulholland happy 😄 

https://github.com/gitify-app/gitify/pull/1268#discussion_r1644892762

bmulholland:
Well, to be fair, I was thinking of going one step further:
`debugging.ts`:
```
function debugLog(msg) {
  // biome-ignore lint/suspicious/noConsoleLog: debug logging
  console.log(msg);
}
```

The point isn't to centralize the biome-ignore, but more to make the intent clear. Specifically: if I'm reading the code, and I see `console.log`, I think "this is a mistake, shouldn't be here." But if I'm reading code and see `debugLog`, I don't question it, and even consider it thoughtful and helpful.

Also, centralizing it means that we could in future e.g. change the debug log to a log file, which could be submitted with bug reporst.

My mental reference here is Rails' log levels: https://guides.rubyonrails.org/debugging_rails_applications.html#sending-messages

Issue #264, opened 2017-02-15
Allow for simple reason filtering


michenly:
Using Github API's reason field
https://developer.github.com/v3/activity/notifications/
<img width="684" alt="screen shot 2017-02-15 at 4 43 02 pm" src="https://cloud.githubusercontent.com/assets/1610169/22996777/eb8a3c7e-f39d-11e6-93af-3a4325578bbe.png">

Do a quick filtering to only show certain type of notification. 
I am thinking a simple toggle of yes or no to included each type of reason.

ie. I can turn off all the notification except for the one specially @mentioned me

icatalina:
It will be cool to have one of this per repo. There are some repos I want to be notified all the time, others where I only care if I'm mentioned....

icatalina:
It will be cool to have one of this per repo. There are some repos I want to be notified all the time, others where I only care if I'm mentioned....

nilsreichardt:
I have the same use case as @icatalina

Juice805:
Isn't this an option in Github itself in the "Watch" Menu? Maybe not back in 2017, but today I think this is already possible.
<img width="359" alt="Screen Shot 2022-05-12 at 4 50 19 PM" src="https://user-images.githubusercontent.com/8647159/168185232-939a473e-f611-4587-beeb-2ff4d0f07706.png">

nilsreichardt:
For me, this not a solution because I subscribed via this menu to many repos. I get ~100 notifications per day because I'm interested in the process of these repositories. But I don't want to get macOS notifications via Gitify for all the 100 GitHub notifications (this would be too crazy and I couldn’t work anymore). It would be nice if I could say that I only want for Repo X or when I'm a participant a Gitify macOS notification.

kpurdon:
This would be a great addition -- github's notification filtering is poor at best, and being able to do even this thin filtering in gitify would be 💯 . 

My specific use case is disabling team mentions so I don't get every PR where the team is left around after auto-assignment to somebody else. There is a setting to disable that, but I don't have control of that.

PR #1273, opened 2024-06-19
feat: group notifications by repository or date

### LINKED ISSUE #836: feature: support grouping notifications by date or repository

setchy:
closes #836.  

for now I've added this as another `Notification Setting` flag. 

Screenshot 2024-06-19 at 6 39 56 AM

---

I've been thinking that in the near future it may make sense to pull out all of our filter-based settings into either their:
- own settings section ie: `Notification Filters`
- dedicated `Filters` page with access from the sidebar  

Types of filter-based settings would include 
- group by repository or date
- show/hide Bot notifications
- show/hide by notification reason #264 
- etc

### DIFF
--- a/src/__mocks__/state-mocks.ts
+++ b/src/__mocks__/state-mocks.ts
@@ -3,6 +3,7 @@ import {
   type AuthState,
   type GitifyState,
   type GitifyUser,
+  GroupBy,
   type Hostname,
   type SettingsState,
   Theme,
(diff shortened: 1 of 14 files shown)


afonsojramos (commented):
I know it is easier to just add a checkbox, but I don't think with a checkbox it is clear what the alternative behaviour is 🤔 Can we use a select instead?

setchy:
> I know it is easier to just add a checkbox, but I don't think with a checkbox it is clear what the alternative behaviour is 🤔 Can we use a select instead?

great feedback.  updated to be a radio group

Screenshot 2024-06-19 at 12 25 17 PM

afonsojramos (approved):
Looking good!! I think this will be one that a good subset of users will enjoy!

Issue #655, opened 2023-10-04
Document non-goals of Gitify


bmulholland:
What things won't we do, and why? Let's write them down.

I'll start by collecting them here. We can add to the README later.

* Operating-system level features
  * Do not disturb, including on schedules. https://github.com/gitify-app/gitify/issues/416#issuecomment-1746480130
  * Persistent notifications like https://github.com/gitify-app/gitify/issues/281. e.g. macOS has Alerts, instead of Banners, which make them persistent
* Seeing past notifications. This is a tool for monitoring new notifications, not seeing old ones, which can be seen at https://github.com/notifications.
* Specific UX/UI changes that add options and/or visual complexity for minor workflow improvements. e.g. https://github.com/gitify-app/gitify/issues/358, https://github.com/gitify-app/gitify/issues/411 and https://github.com/gitify-app/gitify/issues/979
* UI for something that isn't core to Gitify, and/or can be trivially done another way. e.g. https://github.com/gitify-app/gitify/issues/476 and https://github.com/gitify-app/gitify/issues/221
* Support anything other than Github. Doing so would be a major undertaking that we maaayyy consider in future, but it would be a ton of work and there's so much tied to Github right now that it's not something I'd want to maintain.

bmulholland:
I've seen this in the README of other projects. WDYT?

setchy:
my 2 cents:  a brief mention in the README with a link to another detailed MD file

PR #1276, opened 2024-06-19
feat: show notification number for issues, prs, discussions

setchy:
**_Updated_**
Screenshot 2024-06-21 at 7 17 43 AM

**_Original_**
Screenshot 2024-06-19 at 9 49 02 AM


afonsojramos (changes requested):
I'd leave this one marinating a bit... I'm not sure if this is useful information and it is polluting an already quite polluted UI.

afonsojramos:
Maybe reducing opacity to like 60? And making it a font step smaller?

setchy:
> I'd leave this one marinating a bit... I'm not sure if this is useful information and it is polluting an already quite polluted UI.

It's consistent with the GitHub UI experience.  Nowhere in our current UX do we mention the issue numbers which seems like a bit of a miss 🙃

setchy:
> Maybe reducing opacity to like 60? And making it a font step smaller?

I have already set it as `xs` instead of `sm`.  Are you suggesting it should be even smaller?

setchy:
Now with updated opacity to 60

Screenshot 2024-06-19 at 11 17 19 AM

setchy:
An alternate idea, we could model this as a another `pill`

Screenshot 2024-06-19 at 11 23 21 AM

setchy:
Another option, adding it below the notification type icon (it's rough, needs proper formatting)

Screenshot 2024-06-19 at 11 39 00 AM

afonsojramos:
> It's another step on our journey to GitHub UI/UX consistency.

I agree with the part about consistency, but we must have less UI elements, because we don't have the luxury of a full page imo

setchy:
> > Maybe reducing opacity to like 60? 
> 
> Now with updated opacity to 60
> 
> Screenshot 2024-06-19 at 11 17 19 AM
> 
> 

This remains as my preferred approach

afonsojramos:
My preferred approach is no issue number at the end of the day. I've locally played around with `xxs` and `opacity-50`, but it all comes down to this: what is the use of the issue number? As something that should focus on the essential, this is definitely not essential. I'd say this even goes against the following non-goal that we defined in #655:

> Specific UX/UI changes that add options and/or visual complexity for minor workflow improvements.

afonsojramos:
@bmulholland @setchy @Araxeus @adufr opinions on the above?

Araxeus:
I personally don't see value in showing issue number in Gitify, but I guess it could be optional?

setchy:
I still think this is pretty innocent change 😅 

image

for short titles, you'll see the number as you do in the other GitHub Notification UIs
for long titles it will overflow off the screen and only be visible on notification mouse-over

afonsojramos:
> If it's such a big concern, another option could be only having it in the mouse-over label text 🤷‍♂️

Point still remains, what is the value of it?

Araxeus:
> I personally use the Issue/PR numbers a lot with the GH CLI

Would you really check out a branch based on a number from a notification title?

I think you would usually enter the notification, see what's actually inside it, then decide if you should check out or not

adufr:
I'll give my opinion on this since I've been tagged, even though I haven't contributed recently

I like the fact that Gitify can be modular. I personally don't have a use case for this, but IMO it's worth adding it behind a setting (probably disabled by default though?)

> we don't have the luxury of a full page imo

@afonsojramos: you are very right, that's why I'd disable it by default, but having multiple features like this that could be enabled/disabled would solve this problem; the user can just add whatever elements he wants and which fits its use-case

@setchy: about the design, I think I prefer this one too 😁 
image

setchy:
Updated with the latest draft proposal

Screenshot 2024-06-21 at 7 17 43 AM

afonsojramos:
> @afonsojramos: you are very right, that's why I'd disable it by default, but having multiple features like this that could be enabled/disabled would solve this problem; the user can just add whatever elements he wants and which fits its use-case

The issue with "the user can just add whatever elements he wants and which fits its use-case", is that then we'll need to have this in mind with future developments, as they are things that we might not be seeing in local development and might break or affect without us realising. Too many options can be very hard to work with. 

**But**, this is minimal I suppose, so I think I'll move forward with this 🚀

setchy:
> However, we'll need another setting for this I reckon

the `detailed notification` setting already controls this ;)

afonsojramos:
> the issue/pr/discussion number is already flagged with the `detailed notification` setting 🙂

@setchy So do colors, last commenter and state, which I think are very useful and with no extra UI elements, while this one does add a new element, hence the ask for a new setting.

setchy:
Added a new setting.

setchy:
Just remembered I need to do some small changes to disable if the detailed notifications setting isn't enabled

src/context/App.tsx
```diff
@@ -62,6 +62,7 @@ export const defaultSettings: SettingsState = {
   showAccountHostname: false,
   delayNotificationState: false,
   showPills: true,
+  showNumber: true,
```
afonsojramos, line 65:
```suggestion
  showNumber: false,
```
Considering the discussion had, I'd say false is more sensible.

afonsojramos (changes requested):

src/context/App.tsx
setchy: I'd rather people opt out like all the other items we show (pills, details notifications)

afonsojramos: I don't fully understand the reasoning here 🤔 Isn't detailed notifications false by default? Same as `ShowAccountHostname`. There is no real consistency here I think, it is just what we define as being important to be default vs what we don't imo.

setchy: we changed that in v5.6.0 - https://github.com/gitify-app/gitify/pull/1128

setchy: `showAccountHostname` was selfishly to help me when rewriting the account module refactoring. It shows when more than 1 account, but for single account users there was no way to control it.  

In time we may make the account header a bit richer, such as:
- profile avatar
- account-specific quick-links.  the current quick-links in the sidebar are a little limited for multiple-account situations.

afonsojramos: Hmmmmmmm, alright then 🤪 Sorry for being a blocker 😇

Issue #979, opened 2024-04-06
Filter notifications by Organization/User


afonsojramos:
### 📝 Provide a description of the new feature

Similar to Github, we could have logos in the sidebar to filter by organization. Yes, it wouldn't have the same value as GitHub, but we already group by repo, so I suppose the value is already there.

<table>
<tr>
 <td>GitHub UI
 <td>Gitify sidebar
<tr>
 <td><img width="285" alt="image" src="https://github.com/gitify-app/gitify/assets/19473034/d8aae46e-b84a-4d34-b059-8ff0754a7006">
 <td><img width="70" alt="image" src="https://github.com/gitify-app/gitify/assets/19473034/f4a1b0aa-38b8-4bbe-a9c1-1bee9624281f">
</table>

setchy:
Now that we have collapsible headers, I wonder if we can achieve something close to this by adding a notification count badge to our headers (account and repository)

setchy:
Something like this (or similar)... 

<img width="498" alt="Screenshot 2024-06-19 at 11 56 09 AM" src="https://github.com/gitify-app/gitify/assets/386277/24d4a3b5-8c26-40b4-9f6d-9571e5060f48">

PR #1278, opened 2024-06-19
feat: add notification count at account and repository level

setchy:
could help with #979 - seeking feedback 🙏

### DIFF
--- a/src/components/AccountNotifications.tsx
+++ b/src/components/AccountNotifications.tsx
@@ -1,13 +1,15 @@
 import {
+  BellIcon,
   ChevronDownIcon,
   ChevronLeftIcon,
   ChevronUpIcon,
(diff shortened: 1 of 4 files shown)


afonsojramos (changes requested):
I'm sorry for trying to block yet another UI change, but I think that, once again, this will add complexity to something that doesn't need the complexity. Tbh, as the person that opened the issue in question, I don't think it makes sense anymore. :(

setchy:
> I'm sorry for trying to block yet another UI change, but I think that, once again, this will add complexity to something that doesn't need the complexity. Tbh, as the person that opened the issue in question, I don't think it makes sense anymore. :( 

All good. Let's close the issue then. 

My goal largely is to get that down to a smaller list. Less is more.

PR #1284, opened 2024-06-20
refactor: move `electron` files into dedicated folder

afonsojramos:


src/electron/main.js
```diff
@@ -8,18 +8,12 @@ const {
...
 require('@electron/remote/main').initialize();
 
-const idleIcon = path.join(
-  __dirname,
-  'assets',
-  'images',
-  'tray-idleTemplate.png',
-);
-const activeIcon = path.join(__dirname, 'assets', 'images', 'tray-active.png');
+const idleIcon = `${__dirname}/../../assets/images/tray-idleTemplate.png`;
```
setchy, line 15: From what I've read about electron, this may not work in the packaged version

setchy:
Oh my yes!!! This has been on my to-do list! Love it!!!! ❤️

src/electron/main.js
setchy: Could be a non issue. Worth validating with the dmg

afonsojramos: Testing now! 🤓

afonsojramos: It did indeed not work, but I'll make a change to not use join nonetheless

afonsojramos: Done! Now working with path.resolve, which is way cleaner than join!

setchy (approved):

PR #1285, opened 2024-06-20
chore: add ts-check to js files

afonsojramos:


setchy (approved):

setchy:
Would love to make these js files TS in time, too

afonsojramos:
> Would love to make these js files TS in time, too

With webpack I think that would bring us more headaches than we need I think, but feel free to explore 🙈

PR #1288, opened 2024-06-20
fix: account header toggle

setchy:
Fix toggle bug introduce during groupBy feature


src/components/AccountNotifications.tsx
```diff
@@ -91,24 +91,26 @@ export const AccountNotifications: FC<IAccountNotifications> = (
         </div>
       )}
 
-      {showAccountNotifications && groupByRepository
-        ? Object.values(groupedNotifications).map((repoNotifications) => {
-            const repoSlug = repoNotifications[0].repository.full_name;
+      {showAccountNotifications
+        ? groupByRepository
```
bmulholland, line 95: As a fix, this is fine, but as code to maintain: I find this hard to reason with. i.e. What do two chained `?` operators do? :)

setchy: Do you find this easier to reason with?

```
{showAccountNotifications && (
  <>
    {groupByRepository
      ? Object.values(groupedNotifications).map((repoNotifications) => {
        const repoSlug = repoNotifications[0].repository.full_name;

        return (
          <RepositoryNotifications
			key={repoSlug}
			repoName={repoSlug}
			repoNotifications={repoNotifications}
		  />
		);
	  })
	: notifications.map((notification) => (
	  <NotificationRow
		key={notification.id}
		notification={notification}
	  />
	))}
  </>
)}
```

setchy: https://github.com/gitify-app/gitify/pull/1306

PR #1292, opened 2024-06-20
refactor: opacity enum

setchy:


src/components/RepositoryNotifications.tsx
```diff
@@ -55,15 +57,20 @@ export const RepositoryNotifications: FC<IRepositoryNotifications> = ({
...
+          )}
+        >
           <AvatarIcon
             title={repoName}
             url={avatarUrl}
             size="medium"
             defaultIcon={MarkGithubIcon}
           />
           <span
-            className="cursor-pointer truncate opacity-90"
```
afonsojramos, line 66: Intentional?

src/types.ts
```diff
@@ -137,6 +137,13 @@ export enum IconColor {
   WHITE = 'text-white',
 }
 
+export enum Opacity {
+  READ = 'opacity-50',
```
afonsojramos, line 141: What's the reasoning behind the name?

src/components/RepositoryNotifications.tsx
```diff
@@ -55,15 +57,20 @@ export const RepositoryNotifications: FC<IRepositoryNotifications> = ({
         className="group flex items-center justify-between bg-gray-100 px-3 py-2 dark:bg-gray-darker dark:text-white"
         onClick={toggleRepositoryNotifications}
       >
-        <div className="mt-0 flex flex-1 space-x-3 overflow-hidden overflow-ellipsis whitespace-nowrap text-sm font-medium">
+        <div
+          className={cn(
+            'flex flex-1 space-x-3 overflow-hidden overflow-ellipsis whitespace-nowrap text-sm font-medium',
+            Opacity.MEDIUM,
```
afonsojramos, line 63: Does the change in opacity look okay in the icon? But I like the consistency with the other change in NotifRow

afonsojramos (approved):

src/types.ts
setchy: That's what we currently set a Notification Row as when it is `READ` during the delay notification state change.

It was what I had used, too, in the draft PR for seeing read notifications.

Do you have a better name suggestion?

src/components/RepositoryNotifications.tsx
setchy: I think it does.  Applying it at the row level seemed to make the opacity hierarchy clearer

src/components/RepositoryNotifications.tsx
setchy: yes, opacity moved up to the div level

PR #1304, opened 2024-06-21
feat: notification filters

### LINKED ISSUE #264: Allow for simple reason filtering

setchy:
New `Filter` sidebar route which contains any post filtering controls, including

- show bot notifications - moved from Settings
- reason filtering - new. closes #264 

Screenshot 2024-06-21 at 10 39 42 AM

When any non-default filters are selected, it will show a visual indicator
Screenshot 2024-06-21 at 10 40 24 AM


src/routes/Filters.test.tsx
```diff
@@ -0,0 +1,240 @@
...
+          >
+            <MemoryRouter>
+              <FiltersRoute />
+            </MemoryRouter>
+          </AppContext.Provider>,
+        );
+      });
+
+      fireEvent.click(screen.getByLabelText('Go Back'));
+      // TODO - add expects fetch notifications to be called.  pending #1305
```
setchy, line 54: Pending: #1305 to do a few final changes in this PR

bmulholland:
I guess the settings store doesn't support storing arrays?

setchy:
> I guess the settings store doesn't support storing arrays?

it can with a bit of work 😅

setchy:
> I guess the settings store doesn't support storing arrays?

updated to use an array `Reason[]` instead of a nasty string 🙈

bmulholland (approved):
Thanks! I think the extra work to make it an array was worth it :)

setchy:
> Thanks! I think the extra work to make it an array was worth it :)

100000%

setchy:
@afonsojramos - looking forward to your feedback on this, too, before I merge

setchy:
small update... changed the `reset to default filters` to match other footer styling

Screenshot 2024-06-28 at 8 18 55 AM

bmulholland:
FWIW, in that last screenshot, there's no explanation text... I was looking for it and it still look me a minute to find the icon. Without the context, I'd never have guessed that's what the button does!

afonsojramos:
What about adding a counter to the filter icon on the sidebar that when hovered showed a tooltip with the filters being used which could be individually removed AND cleared?

src/routes/Filters.tsx
```diff
@@ -0,0 +1,110 @@
...
+    <div className="flex h-screen flex-col" data-testid="filters">
+      <Header fetchOnBack={true}>Filters</Header>
+      <div className="flex-grow overflow-x-auto px-8">
+        <fieldset className="mb-3">
+          <legend id="notifications" className="mb-1 mt-2 font-semibold">
+            Users
+          </legend>
+          <Checkbox
+            name="showBots"
+            label="Show notifications from Bot accounts"
```
afonsojramos, line 45: What if we change this into a "Hide notifications from Bot Accounts" instead? This way all filters would be disabled by default.

setchy: Good suggestion. I'll make that change shortly

setchy: Renamed the setting and updated labels as suggested

afonsojramos (approved):
Looks awesome!! 🚀 Excited for the next release 👀

PR #1305, opened 2024-06-21
feat: fetch notifications on navigation

setchy:
Currently `fetchNotifcations` is called each time when a subset of settings are changed.  

This can cause rate limiting if enough settings are changed in quick succession.

Instead, this PR moves the retrieve new notifications to occur when moving back to the Notifications screen, either by;

- clicking the `Go Back` button in the header
- clicking on the Sidebar `Settings` icon


afonsojramos (approved):
Looking good! A definite improvement!

PR #1306, opened 2024-06-21
refactor: remove null ternaries

setchy:


afonsojramos (approved):

bmulholland:
I do find it a bit more readable :)

One second thought on my feedback previously is that I mostly work with Vue. There may be React-standard things that are easy to read for those well-versed in that ecosystem. In those case, maybe my feedback isn't valid :)

PR #1320, opened 2024-06-28
refactor: split settings into smaller components

setchy:
Split our growing `Settings` component into smaller, more focused components.

- Appearance Settings
- Notification Settings
- System Settings
- Settings Footer


bmulholland (approved):
It's a nice cleanup, thank you!

I do wonder if these per-component test files are more effort than they're worth. Integration-level tests could cover all of these, pretty much, without needing to have a test-per. As it is, I feel that the tests make the system harder to change without contributing much resiliency to errors.

setchy:
> I do wonder if these per-component test files are more effort than they're worth. Integration-level tests could cover all of these, pretty much, without needing to have a test-per. As it is, I feel that the tests make the system harder to change without contributing much resiliency to errors.

I would argue the opposite. They're easy to write and easy to make changes. I haven't had any concerns

bmulholland:
Oh cool, good news!

afonsojramos:
It would have been way easier to merge this after #1318 :(

afonsojramos:
Nevermind, it was rather easy afterall 😁

PR #1321, opened 2024-06-28
feat: reset to default settings

setchy:
Add a `reset default settings` option.  Of increasing usefulness as we've been adding, and may continue to add more settings in future.

Screenshot 2024-06-28 at 8 14 28 AM


bmulholland:
I perceive that icon as "undo my last change," not "undo all changes, ever." I could, therefore, click it and get effective data loss without intending...

setchy:
Is there an alternative icon you'd propose - https://primer.style/foundations/icons?

Would there be a different location you'd prefer? 

I figured the hover text was enough 🤷‍♂️

afonsojramos:
I'd argue more against the feature, than the icon itself. Why would we want this?

(But if we really want this, I'd say actual text aligned to the right saying "Reset to defaults \*currentUndoIconHere\*")

setchy:
> I'd argue more against the feature, than the icon itself. Why would we want this?

I was a little unclear in my PR.  I think this could be helpful for users who have ignored new settings, or, tinkered with settings and want to revert back to the out-of-the-box / default combination of settings that we recommend.

> (But if we really want this, I'd say actual text aligned to the right saying "Reset to defaults *currentUndoIconHere*")

How about this 😎 - update pushed to branch

Screenshot 2024-07-03 at 6 08 25 PM

afonsojramos:
@setchy took the liberty to push to this branch the option I think is the most viable. Hope that's okay.

Let me know what you think!
image

setchy:
Works for me

afonsojramos:
I think it is a bit more clear right out of the gate and it's only visible if you're looking for it. Which, for a "deep" setting, makes sense!

afonsojramos (approved):

PR #1324, opened 2024-07-01
chore: automatically open devtools on local development

afonsojramos:
I'd argue the "old" settings were not really useful, so I've also removed them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,8 @@ To run the **electron app**:
 pnpm start
 ```
 
+To reload the app with the changes that `pnpm watch` has detected, you can use the `CmdOrCtrl+R` shortcut.
+
 ### Tests
 
 There are 2 checks:
--- a/src/electron/main.js
+++ b/src/electron/main.js
@@ -35,16 +35,12 @@ const browserWindowOpts = {
 
 const contextMenu = Menu.buildFromTemplate([
   {
-    label: 'View',
-    submenu: [
-      { role: 'reload' },
-      { role: 'toggleDevTools' },
-      { type: 'separator' },
-      { role: 'resetZoom' },
-      { role: 'zoomIn' },
-      { role: 'zoomOut' },
-    ],
+    role: 'reload',
   },
+  {
+    role: 'toggleDevTools',
+  },
+  { type: 'separator' },
   {
     label: 'Quit',
     click: () => {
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -30,6 +30,8 @@ To run the **electron app**:
 pnpm start
 ```
 
+To reload the app with the changes that `pnpm watch` has detected, you can use the `CmdOrCtrl+R` shortcut.
```
setchy, line 33: Handy tip 😎

src/electron/main.js
```diff
@@ -77,16 +77,13 @@ app.whenReady().then(async () => {
 
     // DevTools configuration
     mb.window.webContents.on('devtools-opened', () => {
```
setchy, line 79: I'm not too sure what type of experience you're looking for with these changes... 

To me, this is less useful DX...  My local dev/debug flow uses Cmd+Opt+I a lot, along with resizing the window so that I can see the Chrome console, network tab or local storage in full.  These changes for me would make it much harder to locally dev/debug

src/electron/main.js
```diff
@@ -77,16 +77,13 @@ app.whenReady().then(async () => {
...
+        mb.window.setAlwaysOnTop(true);
     });
 
-    mb.window.webContents.on('devtools-closed', () => {
-      const trayBounds = mb.tray.getBounds();
-      mb.window.setSize(browserWindowOpts.width, browserWindowOpts.height);
-      mb.positioner.move('trayCenter', trayBounds);
-      mb.window.resizable = false;
+    mb.window.on('ready-to-show', () => {
+      if (process.env.NODE_ENV === 'development')
```
setchy, line 85: not sure why, but for me when running this locally it didn't auto-show 🤷‍♂️

setchy (changes requested):

src/electron/main.js
afonsojramos: Ah, I see! I use the devtools detached on a new window, but what you are saying makes sense for you usage style.

afonsojramos: Maybe a solution with both might be useful to everyone then, aka, just adding the setAlwaysOnTop?

src/electron/main.js
afonsojramos: Gone!

setchy (approved):
might be nice to only show the dev context menu items (reload, toggledevtools) if running locally, but otherwise lgtm!

afonsojramos:
We're all developers here, so why not 🤠 Maybe it sparks interest in someone contributing 🤞

PR #1330, opened 2024-07-04
feat(sidebar): highlight selected route

setchy:
Highlight sidebar button when route being shown


bmulholland:
I expected to love it, but found the video confusing. I think it's because only two icons change color? Also, something about the red throws me off, like it's an error state.

That could just be because I'm not the one controlling it, though. So may or may not be valid feedback :)

bmulholland (approved):
Code looks good :)

setchy:
> I expected to love it, but found the video confusing. I think it's because only two icons change color? Also, something about the red throws me off, like it's an error state.
> 
> That could just be because I'm not the one controlling it, though. So may or may not be valid feedback :)

It's meant to be orange, but agree it does come across differently.  Is there a different color you think would work well?

afonsojramos:
Maybe no color, but a little triangle next to it? Let me draw something up!

afonsojramos:
Like this? 
image

Otherwise, I'd just stick with the color on hover.

setchy:
Or using a dot (ignore formatting/alignment for now) with matching icon color

Screenshot 2024-07-08 at 11 10 32 AM
Screenshot 2024-07-08 at 11 10 37 AM

afonsojramos:
image
This is too packed. I say let's go with the hover's opacity, simpler implementation and cleaner UI

bmulholland:
Yeah I agree that moving the icon would be weird.

setchy:
agree, didn't intend for the icons to move around in the final version, hence the `(ignore formatting/alignment for now)` comment.

Nonetheless, will look to update the hover opacity instead

setchy:
Me again 😅 

How about a simple tailwind ring?

Screenshot 2024-07-09 at 9 24 27 AM
Screenshot 2024-07-09 at 9 24 34 AM

bmulholland:
I like the ring. Personal opinion though.

setchy:
Updated to use a shadow. imho, this is the best of the options explored above. 


https://github.com/user-attachments/assets/a443b79d-30e6-42b5-94a5-e6862a95ce8b

setchy:
Moving back to draft as @afonsojramos is looking into some things on his side

setchy:
Closing this concept for now - I'll raise a separate PR just for updating the accounts route

PR #1362, opened 2024-07-11
build: revert to use pnpx

setchy:
see https://github.com/gitify-app/gitify/pull/1357#issuecomment-2221751031

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-pnpm lint-staged
+pnpx lint-staged


afonsojramos:
Obviously that pnpx works, but you should check your config first here, because it should work 👀

afonsojramos:
But to be honest, shouldn't be much of an issue. And lint-staged actually recommends this config, so I'll excuse myself here

afonsojramos (approved):

PR #1376, opened 2024-07-13
feat(settings): open links in foreground or background

setchy:


afonsojramos:
I suppose here is where @bmulholland's point comes into play the most. Why not just leave this as a cmd + click action, and remove the setting altogether?

bmulholland (approved):

bmulholland:
I still don't mind Cmd+Click, but do wonder about discoverability.

afonsojramos (approved):

PR #1389, opened 2024-07-17
feat: mark notification as read to allow transition effects time

setchy:
Currently, there is no delay to allow the transition effect time to perform.  Switching back to an explicit "mark as read" buys us some time and now displays the notification transition effect


bmulholland:
Plus:
Screenshot 2024-07-17 at 14 45 35

bmulholland (approved):

Issue #1395, opened 2024-07-21
Check for new versions of Gitify


bquorning:
### 📝 Provide a description of the new feature

I am always a couple of months behind the latest version, because I forget to check for updates to Gitify. Are there any plans for adding an automatic “check for new releases” feature?

setchy:
We do already use [electron-updater](https://www.npmjs.com/package/electron-updater) to perform [automated update checks](https://github.com/gitify-app/gitify/blob/f513aa3441a8f4af84beababb3dff3d6f40ff930/src/electron/main.js#L64)

bquorning:
I am using macOS Sonoma 14.5.

If the check only happens when Gitify is started (seeing `app.whenReady().then` and `on('ready'` in the code) then maybe it’s because I don’t restart my computer (or Gitify) very often.

Of course, it could also just be that I’m haven’t noticed the notifications 😅

Before manually updating Gitify next time, I’ll make sure to quit and restart it to see if I get a notification – and if not, I’ll reopen this issue.

PR #1397, opened 2024-07-21
feat: check for updates

setchy:
Based on #1395 from @bquorning 

Draft PR to consider:
- Adding a context menu for checking for updates manually
- Adding a scheduled timer to check for updates automatically (not just on app start)
- Display somewhere in the UI a visual indicator when there is a new version available


setchy:
**Latest version**
Screenshot 2024-07-22 at 10 30 23 AM

**Update available**
Screenshot 2024-07-22 at 10 31 19 AM

afonsojramos:
Love the idea, but I think the implementation in the UI is a bit weird. The behaviour that I see in these menubar apps is usually an indicator in the icon of the menubar and I think it is less intrusive and more informative at the same time, what do you think?

Example: Ollama
image
image
image

setchy:
> Love the idea, but I think the implementation in the UI is a bit weird. The behaviour that I see in these menubar apps is usually an indicator in the icon of the menubar and I think it is less intrusive and more informative at the same time, what do you think?
> 
> Example: Ollama
> image
> image
> image
> 

I like that approach. Would you be interested in helping pivot this way?

setchy:
I also like the `restart to update` context menu entry. That seems better than doing it automatically

setchy:
Another thought, maybe our menu system could adopt something like this

- check for update 
-- show status if update available 
-- show restart to update if available 
- separator
- Developer (sub menu)
-- Reload
-- Toggle Developer Tools
-- separator 
- Quit

setchy:
@afonsojramos - pushed updates to this branch that add the extra context menu elements.

Not sure the best way to update the tray icon.  We'd either need to create a new png, or, manipulate the tray title (used for notification counts) to display a visual indicator

bmulholland (approved):
Really nice

PR #1411, opened 2024-07-25
feat: use first account for quick link hostname

setchy:
closes https://github.com/gitify-app/gitify/issues/1410

### DIFF
--- a/src/utils/auth/migration.ts
+++ b/src/utils/auth/migration.ts
@@ -13,7 +13,6 @@ export async function migrateAuthenticatedAccounts() {
   const existing = loadState();
 
   if (!hasAccountsToMigrate(existing.auth)) {
-    log.info('Account Migration: No accounts need migrating');
     return;
   }
 
--- a/src/utils/links.ts
+++ b/src/utils/links.ts
@@ -15,20 +15,20 @@ export function openGitifyReleaseNotes(version: string) {
   );
 }
 
-export function openGitHubNotifications(account: Account) {
-  const url = new URL(`https://${account.hostname}`);
+export function openGitHubNotifications(hostname: Hostname) {
+  const url = new URL(`https://${hostname}`);
   url.pathname = 'notifications';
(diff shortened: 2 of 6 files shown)


src/utils/auth/migration.ts
```diff
@@ -13,7 +13,6 @@ export async function migrateAuthenticatedAccounts() {
   const existing = loadState();
 
   if (!hasAccountsToMigrate(existing.auth)) {
-    log.info('Account Migration: No accounts need migrating');
```
setchy, line 16: was making our jest console busy

src/utils/links.ts
```diff
@@ -15,16 +15,22 @@ export function openGitifyReleaseNotes(version: string) {
   );
 }
 
-export function openGitHubNotifications() {
-  openExternalLink('https://github.com/notifications' as Link);
+export function openGitHubNotifications(account: Account) {
```
akellbl4, line 18: Maybe a good idea to default the hostname to `github.com`, I'm not sure if the sidebar is available while no accounts are present.

For simplicity of use, the link handlers can be grouped into a custom hook and pull needed context on their own.

src/utils/links.ts
```diff
@@ -15,16 +15,22 @@ export function openGitifyReleaseNotes(version: string) {
...
+  const url = new URL(`https://${account.hostname}`);
+  url.pathname = 'notifications';
+  openExternalLink(url.toString() as Link);
```
akellbl4, line 21: As I can see all use cases of `openExternalLink` pass string as a parameter. So I would say that probably it should accept the argument as `string` since all external URLs are not created as a branded `Link`

akellbl4 (commented):
Thanks for quickly implementing the feature. I've left some comments in case you find them valuable.

src/utils/links.ts
setchy: Great feedback: updated to handle the unauthenticated scenario

PR #1420, opened 2024-07-28
refactor: fix biomejs lint warning

setchy:


bmulholland (approved):

src/context/App.tsx
```diff
@@ -173,7 +173,7 @@ export const AppProvider = ({ children }: { children: ReactNode }) => {
     const newSettings = { ...settings, ...defaultFilters };
     setSettings(newSettings);
     saveState({ auth, settings: newSettings });
-  }, [auth]);
+  }, [auth, settings]);
```
afonsojramos, line 176: Be careful with this 👀 All linters are known for sometimes inducing people in error with this, for example in useEffects

setchy: Thanks. So for this instance do you think this should be reverted?

PR #1423, opened 2024-07-30
feat: add conditional render on "Mark as Done" Btn

julienperrault:
Entreprise API doesn't provide "Mark as Done". So i have done this PR to remove "Mark as Done" Btns.
closes : https://github.com/gitify-app/gitify/issues/1412

### DIFF
--- a/src/components/NotificationRow.tsx
+++ b/src/components/NotificationRow.tsx
@@ -10,7 +10,7 @@ import { AppContext } from '../context/App';
 import { Opacity, Size } from '../types';
 import type { Notification } from '../typesGitHub';
 import { cn } from '../utils/cn';
-import { formatForDisplay } from '../utils/helpers';
+import { formatForDisplay, isEnterpriseServerHost } from '../utils/helpers';
 import {
   getNotificationTypeIcon,
   getNotificationTypeIconColor,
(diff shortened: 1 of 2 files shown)


setchy (changes requested):
"Mark as Done" was added to GitHub Enterprise Server in version 3.13. See https://docs.github.com/en/enterprise-server@3.13/rest/activity/notifications?apiVersion=2022-11-28#mark-a-thread-as-done

Instead, we should look to make the API call resilient for users on GitHub Enterprise Servier 3.12 or older

setchy:
@julienperrault - thoughts on https://github.com/gitify-app/gitify/pull/1424 👀

akellbl4:
@setchy I like the solution with hiding, but it can be improved by checking the GHE version and not hiding for any GHE instance.

akellbl4:
I haven't checked but if the app switches to https://github.com/octokit/octokit.js it might be easier to detect supported features

setchy:
> @setchy I like the solution with hiding, but it can be improved by checking the GHE version and not hiding for any GHE instance.

agree :) - updated https://github.com/gitify-app/gitify/pull/1424 to use version information.  

If you are able to test the branch locally @akellbl4 @julienperrault that would be greatly appreciated 🙏

setchy:
superseded by https://github.com/gitify-app/gitify/pull/1424

PR #1439, opened 2024-08-06
feat(settings): disable `show accounts` when multiple accounts are authenticated

setchy:
Toggling this setting has no affect when there are multiple accounts authenticated.


src/utils/auth/utils.ts
```diff
@@ -218,3 +218,11 @@ export function isValidToken(token: Token) {
 export function getAccountUUID(account: Account): string {
   return btoa(`${account.hostname}-${account.user.id}-${account.method}`);
 }
+
+export function hasAccounts(auth: AuthState) {
```
afonsojramos, line 222: 😅

src/components/settings/AppearanceSettings.tsx
```diff
@@ -198,7 +199,8 @@ export const AppearanceSettings: FC = () => {
       <Checkbox
         name="showAccountHostname"
         label="Show account hostname"
-        checked={settings.showAccountHostname}
+        checked={settings.showAccountHostname || hasMultipleAccounts(auth)}
+        disabled={hasMultipleAccounts(auth)}
```
afonsojramos, line 203: Why exactly do we need to disable this on multiple accounts?

afonsojramos: I see, but can't the hostname be the same on some situations?

setchy: Indeed, we support repeated hostnames now.

setchy: so single account, feature flag toggles showing the account header
multiple accounts, feature flag does nada

afonsojramos: Right, but does it make sense for it to be disabled on the situations where it is the same for example?

setchy: `showAccountHostname` is misleading nowdays.  It really should be `showAccountNameHeader` or something like that as it represents the row containing account name, account avatar, account platform type, etc

PR #1449, opened 2024-08-08
fix: cross-platform emoji compatibility

### LINKED ISSUE #1446: review use of emojis for cross-platform compatability
_Originally posted by @Araxeus in https://github.com/gitify-app/gitify/issues/1441#issuecomment-2273613028_

> 
> Same but the network error icon isnt even rendering for me (Windows)
> 
> image
> 

This issue is to review all of our emojis to ensure they exist on macOS, Windows and Linux

setchy:
Resolves #1446

Use the [@discordapp/twemoji](https://github.com/discord/twemoji) (a fork of twemoji which supports locally packaged SVGs) to convert emojis into SVGs for cross-platform compatibility.

twemoji supports the latest [Unicode Emoji 15.1](https://unicode.org/emoji/charts-15.1/emoji-released.html)


setchy:
@Araxeus - if you happy to have a few minutes, I'd really appreciate to test this branch locally and confirm that this fixes the emoji compatibility issues on Windows 10.  If not, that's OK too 🙂

Araxeus:
How do I easily test the changes? im not sure how to display them in the app?

setchy:
> How do I easily test the changes? im not sure how to display them in the app? they are locked behind certain conditions right?

Turning off WiFi should trigger the `Network Connectivity` error page which you previously had a broken emoji for

Araxeus:
image

doesn;t seem to work on my Windows machine, does it actually work on your pc?

setchy:
Apologies for that @Araxeus - pushed a fix.  Didn't think that when offline, the app wouldn't be able to fetch the SVG from the CDN 🤦.

Updated to handle offline SVGs for network connectivity loss

src/components/icons/Emoji.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type TwemojiOptions = {
+  base: string;
+  size: string;
+  ext: string;
+};
+
+export const Emoji: FC<IEmoji> = (props: IEmoji) => {
+  // Generate the HTML for the emoji using twemoji
+  const emojiHtml = twemoji.parse(props.emoji, {
```
setchy, line 17: We could also decide to store ALL emoji svgs locally instead of conditionally routing based on the callback fn below

Araxeus:
I've only just realized they are fetched from cdn... personally I'm not a big fan of this approach - emojis are simple immutable things which could all just be stored locally

Those svg files could even be included in an npm dependency if needed

But i might be a minority with this opinion 🤷‍♂️

setchy:
Thanks for retesting @Araxeus and confirming it works now.

> I've only just realized they are fetched from cdn... personally I'm not a big fan of this approach - emojis are simple immutable things which could all just be stored locally
I agree, this was my initial reaction too. But after some more reading it seems this is a popular approach for x-platform compatibility and used in platforms like Twitter/X (naturally), Discord etc.  Once fetched the first time the assets will be cached in the browser.  

I did consider storing all of the emojis we use as local SVG assets, but was uneasy with the thought of having to maintain them if/as the Unicode Emoji spec evolves.

> Those svg files could even be included in an npm dependency if needed
I did look around for a while but couldn't find any maintained packages for twemoji svgs (eg: https://www.npmjs.com/package/@twemoji/svg)

Araxeus:
> I did look around for a while but couldn't find any maintained packages for twemoji svgs (eg: [npmjs.com/package/@twemoji/svg](https://www.npmjs.com/package/@twemoji/svg))

should be pretty simple to remake this library, i dont know why the author turned off auto updates but it should be kept in sync via github actions - https://github.com/boywithkeyboard-archive/twemoji_svg/blob/b5b7226b48d30c215d45f98cbc55691b616ab685/.github/workflows/update.yml

https://github.com/boywithkeyboard-archive/twemoji_svg/blob/main/update.ts

I could remake this package this weekend

setchy:
Left a comment on https://github.com/jdecked/twemoji/issues/58.  The discord fork https://github.com/discord/twemoji does package SVGs, so perhaps that's an interim route...

src/components/icons/Emoji.tsx
setchy: This is no longer a concern after pivoting to the https://github.com/discord/twemoji fork which contains SVGs with `/dist`

setchy:
Pivoted to https://github.com/discord/twemoji to source all emojis locally from `node_modules`

Araxeus:
Is it not better to individually select emoji's to enable tree shaking? (so that we bundle only the required emojis)

afonsojramos:
> Is it not better to individually select emoji's to enable tree shaking? (so that we bundle only the required emojis)
> 
> not super important unless all the svgs weight alot

Agree with the above!

setchy:
the increased app size with this PR is between 1-2 MB depending on platform.  to me, the proposed solution here is adequate enough for what we're trying to solve.

Araxeus:
I just don't think that packaging the whole svg library serves anything except convenience for the devs

the end users of the app doesn't actually need all those files, for them it's just bloat (it's not really important but could be avoided IMO)

tbh this is really more of a philosophical issue, when you consider the app is already 317MB! of electron bloat on my pc - it doesn't add alot

bmulholland:
@Araxeus If it's indeed 317MB, and the icons are 2MB, then this is 0.6%. From what I understand, e.g. the exe goes from 83->84MB, which is 1.2%. If app size was a concern, which TBH for me it's not, then there's almost certainly bigger gains to be made elsewhere.

If you want to talk about overall app size, "electron bloat," etc, can I suggest a discussion focused on that, perhaps as a dedicated issue? We'd first need to align on prioritization between developer experience and app size. And the impact of the app size and memory usage vs that tradeoff.

Araxeus:
You could also add 1000 lines of text to the app and only use one of those lines

The app size would go up by less than 1mb 🤷‍♂️

----

Honestly i feel like the "bad guy" arguing about this, I wrote like 20 comments on this PR and im not even part of @gitify-app and this **isnt important at all** - whatever you end up deciding will be fine by me (I wont argue about this anymore)

Thank you everyone for the hard work on the app, especially @setchy for the work on this PR 💪

setchy:
@Araxeus - certainly not the "bad guy" at all - I sincerely appreciate the engagement, feedback and discussion.  Huge thank you 🙏

afonsojramos (commented):
Let's ship it! 🚀

afonsojramos (approved):
Let's ship it! 🚀

PR #1451, opened 2024-08-09
fix: openLinks on first login screen

shiron-dev:
There was a bug that the "Create new OAuth App" button did not respond when pressing it to login again after deleting data.

This happens because loadState() is executed when opening the link, but once the data is deleted, localStorage becomes empty and loadState() returns null.
So I fixed this bug by saving the defalut states (defaultAuth and defaultSettings) when localStorage is empty.

After starting the app with `pnpm start` and then pressing the "Create new OAuth App" button, the error displayed was as follows
```
[44950:0809/151522.895514:INFO:CONSOLE(2)] "Uncaught TypeError: Cannot read properties of undefined (reading 'openLinks')", source: file:///Users/shiron/projects/gitify/build/js/app.js (2)
```

### DIFF
--- a/src/context/App.tsx
+++ b/src/context/App.tsx
@@ -46,7 +46,7 @@ import { clearState, loadState, saveState } from '../utils/storage';
 import { setTheme } from '../utils/theme';
 import { zoomPercentageToLevel } from '../utils/zoom';
 
-const defaultAuth: AuthState = {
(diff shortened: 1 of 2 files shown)


setchy:
Thanks @shiron-dev for reporting this and raising a fix.

I've taken your approach and moved it into our storage `loadState` fn

PR #1464, opened 2024-08-12
refactor: rename `showAccountHostname` to `showAccountHeader`

setchy:
From our simple beginnings as a `hostname` only account header, we have grown to now show avatar, login name, account-specific quick links and more. 

As such, it's time to rename these vars.

The setting rename is somewhat breaking, but given it's a setting that isn't that commonly used (assumption), i think it's OK


bmulholland:
Isn't there a way to migrate existing setting state? We could add code to port over the setting, if set, and keep it in for a few releases.

setchy:
> Isn't there a way to migrate existing setting state? We could add code to port over the setting, if set, and keep it in for a few releases.

we have account migration logic, but not setting atm.  i think we should be ok this time :)

bmulholland (approved):

PR #1476, opened 2024-08-19
build: prepare `v5.13.0` release

setchy:
https://github.com/gitify-app/gitify/releases/tag/v5.13.0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - 'release/v**'
   pull_request:
     branches:
       - main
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - 'release/v**'
```
setchy, line 7: I don't see a need for us to do both a regular build, and a release build on `release/v***` branches

afonsojramos: I suppose if one depended on the other, it would make sense, but not like this. However, pretty sure the builds are required in the PR before merging... Nothing we can't override obviously

afonsojramos (approved):

.github/workflows/build.yml
afonsojramos: Nevermind, this is just for the push event, merge away!

PR #1498, opened 2024-08-30
feat: mark as done on unsubscribe

### LINKED ISSUE #1497: feat: allow `mark as done` on unsubscribe
clicking the bell (unsubscribe) doesn't "mark as done" in a batch (but i worked the habit of 2 clicks)

_Originally posted by @y-nk in https://github.com/gitify-app/gitify/issues/1485#issuecomment-2319702596_

setchy:
closes #1497 

Screenshot 2024-08-30 at 8 16 20 AM


bmulholland (approved):
Why not do this by default?

setchy:
> Why not do this by default?

No real reason... 

For comparison, "mark as done on open" is opt-in. 

The good thing is we can always update the defaults in future

bmulholland:
> For comparison, "mark as done on open" is opt-in.

IIRC that was partly because Github usually does that itself. Could be wrong.

The thing that's even more so about this action is that unsubscribe is a write action already; mark as done seems usually implicit to me.

Just my 2c :)

y-nk:
extremely grateful for this 🙏

PR #1506, opened 2024-09-02
refactor: extract error constants

setchy:
goal is to slowly working towards separating the gitify specific constants/types from the github (or other platform) constants/types


bmulholland (approved):
Nice cleanup :)

PR #1527, opened 2024-09-14
fix: restore settings order

setchy:
Upon first start, filters were not loaded before the first notification fetch occurred resulting in an inconsistent state.

This fixes it by simply restoring settings (and filters) before restoring accounts


bmulholland (approved):
Nice find

PR #1537, opened 2024-09-17
refactor: change chevron type for closed sections

setchy:
Use a chevron right icon when the headers are collapsed.  More intuitive UX.


bmulholland (approved):
Yeah this is better for sure. Thanks!

PR #1538, opened 2024-09-17
refactor: update sidebar labels

setchy:
Change casing used for sidebar labels


bmulholland (approved):
So we're standardizing on sentence case?

setchy:
> So we're standardizing on sentence case?

I think so. Looks a little better imho

PR #1560, opened 2024-09-27
refactor: combine mark as fns to support a list of notifications

setchy:
Consolidate our remove, markAsRead and markAsDone logic to support an array of notifications.


bmulholland (approved):
Always love a good cleanup

PR #1564, opened 2024-09-30
refactor: major project structure update

setchy:
Major project structure update applying
- learnings from building [atlassify](https://github.com/setchy/atlassify)
- converge closer to the latest https://electron-react-boilerplate.js.org/ example

Main updates include:
* split code into `main` and `renderer` src components for clear separation of duties
* update `main` to use TypeScript and split into smaller files (main, menu, icons, utils, updater)
* improve webpack configs, including code minimizing for production builds
* use webpack plugins for templating html and extracting css
* use webpack plugins to copy twemoji svgs into final build with updated EmojiText implementation to avoid reading directly from node_modules path.
* move all non-core dependencies from `dependencies` to `devDependencies`
* rename jest test suites to match new folder structure


src/renderer/components/EmojiText.tsx
```diff
@@ -21,13 +21,12 @@ export const EmojiText: FC<IEmojiText> = ({ text }) => {
       ref.current.innerHTML = twemoji.parse(text, {
         folder: 'svg',
         ext: '.svg',
-        callback: (icon: string, options: TwemojiOptions, _variant: string) => {
```
setchy, line 24: No longer need to pluck these out from the node_modules within the asar bundle 🚀

bmulholland:
Any way to split this into chunks, to ease review? :)

setchy:
> Any way to split this into chunks, to ease review? :)

Yep - can be done...

.github/workflows/release.yml
```diff
@@ -82,9 +84,10 @@ jobs:
         env:
           OAUTH_CLIENT_ID: ${{ secrets.oauth_client_id }}
           OAUTH_CLIENT_SECRET: ${{ secrets.oauth_client_secret }}
-      - run: pnpm make:linux --publish onTagOrDraft
+      - run: pnpm package:linux --publish onTagOrDraft
         env:
           GH_TOKEN: ${{ secrets.github_token }}
+      - run: pnpm prepare:remove-source-maps
```
afonsojramos, line 90: 1. Why do we need to remove the source-maps?
2. Why is this one after the package and on others before the package?

setchy: Good eyes re: 2 - that's an oversight on my part.  Updated

setchy: re 1 - it was within the https://github.com/electron-react-boilerplate/electron-react-boilerplate example.  My assumption that they remove the source maps so that the packaged+minified code cannot be reconstructed.  Not a big deal for an open source project

src/renderer/utils/cn.ts
```diff
@@ -1,4 +1,4 @@
-import { type ClassValue, clsx } from 'clsx/lite';
+import { type ClassValue, clsx } from 'clsx';
```
afonsojramos, line 1: Did you find a reason for needing the non lite one? I had set that one because we didn't need the base one.

config/webpack.config.main.prod.ts
```diff
@@ -0,0 +1,17 @@
+import TerserPlugin from 'terser-webpack-plugin';
+import type webpack from 'webpack';
+import { merge } from 'webpack-merge';
+import baseConfig from './webpack.config.main.base';
+
+const configuration: webpack.Configuration = {
+  devtool: 'source-map',
```
afonsojramos, line 7: Do we need this in the prod one?

afonsojramos (changes requested):
Looking good! A few comments, but I think this may make it easier for us to switch from Webpack to Vite in the future 👀

.github/workflows/release.yml
afonsojramos: Exactly, and it may help with tracing back bugs reported by users, however I am not sure how we can help users provide these off the top of my head 🤔

setchy: Yeah, not too sure.  I think our exception handling and electron-logger should be catching most user reported issues.

That said, would it be best for now to skip minimizing source code?

src/renderer/utils/cn.ts
setchy: No use-case, but was having issues with the new tsconfig and importing clsx/lite.

config/webpack.config.main.prod.ts
setchy: 🤷‍♂️

package.json
```diff
@@ -95,58 +114,67 @@
...
     "menubar": "9.5.0",
-    "nprogress": "0.2.0",
     "react": "18.3.1",
     "react-dom": "18.3.1",
-    "react-final-form": "6.5.9",
     "react-router-dom": "6.26.2",
-    "tailwind-merge": "2.5.2",
-    "ts-loader": "9.5.1",
-    "typescript": "5.6.2",
     "update-electron-app": "3.0.0"
```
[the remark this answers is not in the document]
setchy: We can validate in the release candidate testing

src/renderer/index.html
bmulholland, line 1: There was a bunch of CSS in the original version of the index, where did that go?

src/renderer/index.tsx
```diff
@@ -2,9 +2,8 @@ import { createRoot } from 'react-dom/client';
 
 import 'nprogress/nprogress.css';
 import 'tailwindcss/tailwind.css';
+import { App } from './App';
 
-import { App } from './app';
-
-const container = document.getElementById('gitify');
+const container = document.getElementById('root');
```
bmulholland, line 7: IMO "root" could be a bit more specific. Root what?

src/renderer/index.html
setchy: That CSS is now homed within `src/renderer/App.css`, imported in `src/renderer/App.tsx` and through webpack plugins packaged into `styles.css` within the output folder

src/renderer/index.tsx
setchy: it's the root div container within the body.  Seems to be commonly used in the templates I've looked at.  

If we feel its better, happy to leave the root div as `gitify`

src/renderer/App.css
bmulholland, line 1: Ah, here it is...

config/webpack.config.main.prod.ts
bmulholland: IMO source-maps make sense if we have a centralized error reporting. That's probably why the template has it. Since we don't have exception tracking, there's nowhere to send the source maps, so I don't think it's necessary.

package.json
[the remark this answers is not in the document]
bmulholland: What's even the difference between the two, in this case? I feel like it makes sense for e.g. a server, where you have execution mode and development mode. But for Gitify, the code that's run in execution mode is all post-build, and the dev dependencies are used to make the build. So the package.json is never actually executed in a non-dev mode. If that makes sense?

bmulholland (approved):

src/renderer/index.tsx
bmulholland: Yeah it's overly picky, all good :)

src/renderer/index.html
bmulholland: Yep, found it later :)

PR #1622, opened 2024-11-05
fix: set 'Cache-Control: no-cache' for GHE server

### LINKED ISSUE #1611: Gitify displays the same notification repeatedly when using GHE Server

- [X] I have checked older issues, open and closed

### 📝 Description

I can't write the details because this problem only occurs in notifications from private GHE server. Sorry.

### Problem

In my environment, notifications that I have already “Marked as read” are displayed every time I refresh Gitify, and other new notifications are not loaded.

### Debugging

When I checked by starting up the Developer Tool, I found that a status code of 205 was being returned for the “Mark as read” process, and `/api/v3/notifications` returned 200. I found that it had already been marked as read in the GHE Server.

I refreshed the Gitify, then this notification appeared again on Gitify. This was the same even if I refreshed the page after marking the notification as read multiple times. Also, in this state, Gitify does not load new notifications from the GHE Server.

However, if I enable “Disable cache” from the Developer Tools, everything will work fine. Gitify will display new notifications and will no longer display notifications that have already been read.If I turn off “Disable cache”, it will return to its original state.

What I found interesting was that when Disable cache was turned off, it looked like no requests were actually being sent when refreshing notifications manually.
Developer Tool shows the message as follows:

スクリーンショット 2024-10-29 14 30 29
When I hover over the mouse cursor, the overlay will display as follows.

> Only provisional headers are available because the request was not sent over the network and instead was served from a local cache, which doesn't store the original request headers. Disable cache to see full request headers.

It seems that Gitify's requests to the GHE Server may include an `If-Modified-Since` header. When I made a request with this header using the `gh api` command, I got the following result. 

```
$ gh api --hostname xxx.xxx.xxx \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  -H 'If-Modified-Since: Tue, 29 Oct 2024 04:57:09 GMT' \
  '/notifications?participating=false'

gh: HTTP 304

$ gh api --hostname xxx.xxx.xxx \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  '/notifications?participating=false'

[]
```

### 🪜 Steps To Reproduce

Use GHE Server.

1. Press "Mark as read" for new notifications
2. Press "Refresh notifications"

### Gitify Version

5.16.1

### Operating System

macOS

### GitHub Account

Combination

### 📸 Screenshots

_No response_

pddg:
fix #1611 

Whether it is `github.com` or GHE Server, set the `Cache-Control: no-cache` header.

### DIFF
--- a/src/renderer/utils/api/__snapshots__/client.test.ts.snap
+++ b/src/renderer/utils/api/__snapshots__/client.test.ts.snap
@@ -22,7 +22,7 @@ exports[`renderer/utils/api/client.ts headNotifications should fetch notificatio
 {
   "Accept": "application/json",
   "Authorization": "token token-123-456",
-  "Cache-Control": "",
+  "Cache-Control": "no-cache",
   "Content-Type": "application/json",
 }
 `;
@@ -76,7 +76,7 @@ exports[`renderer/utils/api/client.ts listNotificationsForAuthenticatedUser shou
 {
   "Accept": "application/json",
   "Authorization": "token 1234568790",
-  "Cache-Control": "",
+  "Cache-Control": "no-cache",
   "Content-Type": "application/json",
 }
 `;
--- a/src/renderer/utils/api/request.ts
+++ b/src/renderer/utils/api/request.ts
@@ -94,6 +94,7 @@ function shouldRequestWithNoCache(url: string) {
 
   switch (parsedUrl.pathname) {
     case '/notifications':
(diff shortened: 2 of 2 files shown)


setchy (approved):
Thank you, great fix!

PR #1692, opened 2024-12-28
refactor(shared): create `electron-log` wrapper

setchy:
Thanks @bmulholland https://github.com/gitify-app/gitify/pull/1681#issuecomment-2564054261

### DIFF
--- a/src/main/utils.ts
+++ b/src/main/utils.ts
@@ -43,7 +43,11 @@ export function openLogsDirectory() {
   const logDirectory = path.dirname(log.transports.file?.getFile()?.path);
 
   if (!logDirectory) {
-    logError('openLogsDirectory', 'Could not find log directory!');
+    logError(
+      'openLogsDirectory',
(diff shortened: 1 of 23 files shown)


setchy:
would like to try and remove the two `shared` -> `render` imports for types and mocks, but otherwise 99% happy with this

src/main/utils.ts
```diff
@@ -41,7 +43,11 @@ export function openLogsDirectory() {
   const logDirectory = path.dirname(log.transports.file?.getFile()?.path);
 
   if (!logDirectory) {
-    log.error('Could not find log directory!');
```
bmulholland, line 44: Looks like it helped with consistently immediately, too :)

bmulholland (commented):
Love it!

src/main/utils.ts
setchy: Absolutely! Great suggestion

PR #1749, opened 2025-01-21
feat: continue primer component and design token migration

setchy:
Two main goals in this PR
- replace *almost all* tailwind colors with GitHub Primer Design Tokens
- Use GitHub Theme Provider to control setting light and dark modes
- ongoing refactor of Gitify UI components to leverage GitHub Primer Components where possible

The only remaining area where we are using `.dark` classes to control theming is our scrollbar (track, thumb and hover) - left a TODO


afonsojramos:
I assume that this PR represents one step towards, but it is not the complete thing yet, right?

setchy:
> I assume that this PR represents one step towards, but it is not the complete thing yet, right?

Good callout, that's correct - a few more steps to get fully there

PR #1757, opened 2025-01-23
fix: prevent default filters from getting overwritten

aravindparappil46:
"Clear Filters" button was not unchecking selected filters because of a reference between `settings` and `defaultSettings`. i.e., whenever filters changed, the filter reasons in the `defaultFilters` object also mutated, which is not desirable.

Using the spread operator to get selected filters fixes this issue.

**Video of bug:**

https://github.com/user-attachments/assets/1719e109-ca10-4c29-afe0-7a3d5337660e


setchy (approved):

setchy:
Thanks @aravindparappil46, appreciate the fix 🛠️

PR #1763, opened 2025-01-23
feat: avatar with fallback icons component

### LINKED ISSUE #1224: Broken avatar on GHE (again)

- [X] I have checked older issues, open and closed

### 📝 Description

In a GHE situation, avatars for users are broken. 
This seems similar to #564 but in fact is not. 

Looking at the inspector tab, the `src` for the image is correctly populated and points a valid image if I open it from my browser.
Incidentally, I have an amount of warnings as 

> reading cookie in cross-site context will be blocked in future Chrome versions

Whether that's the real problem or not...I am yet to understand. I cannot find any other error.

### 🪜 Steps To Reproduce

1. Use in a GHE context
2. GHE is 3.11

### Gitify Version

5.7.0

### Operating System

macOS

### GitHub Account

GitHub Enterprise

### 📸 Screenshots

Screenshot 2024-06-12 at 10 39 10

setchy:
New avatar component that handles img error loading with fallback icons

| 😄    | 😅 |
| -------- | ------- |
| 2025-01-23T12-19-22 379Z-Gitify-screenshot |  2025-01-23T12-18-50 648Z-Gitify-screenshot  |

Closes #1224 (or at least provides a "better" option than completely broken images)

### DIFF
--- a/src/renderer/components/avatars/AvatarWithFallback.tsx
+++ b/src/renderer/components/avatars/AvatarWithFallback.tsx
@@ -5,13 +5,15 @@ import { FeedPersonIcon, MarkGithubIcon } from '@primer/octicons-react';
 import { Avatar, Stack, Text } from '@primer/react';
 
 import { type Link, Size } from '../../types';
+import type { UserType } from '../../typesGitHub';
+import { isNonHumanUser } from '../../utils/helpers';
 
 export interface IAvatarWithFallback {
-  src: Link;
-  alt: string;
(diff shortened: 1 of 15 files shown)


src/renderer/components/avatars/AvatarWithFallback.tsx
bmulholland, line 1: I'm a huge fan of components like these -- nice work!

PR #1771, opened 2025-01-24
chore(deps): update tailwindcss to v4

bmulholland:
This was pretty much just the auto-upgrader. I haven't done significant testing here. yet, just a quick pass.

tailwind 4 includes a change to do config as CSS, which I chose to split out into a separate step.


assets/css/main.css
```diff
@@ -1,3 +1,21 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import "tailwindcss";
```
setchy, line 1: would this also work in our `./src/renderer/App.css` ?

bmulholland: Oh, yeah, totally. That's where it should go actually.

bmulholland: I'm out of time for this today, feel free to make that change directly. Or I'll do next week.

setchy (approved):
Thanks @bmulholland - this looks great

PR #1772, opened 2025-01-24
chore: Switch to Tailwind v4 CSS config

bmulholland:
This is their new configuration format, but it's a bit tricky. I'm not confident all these changes are working as expected -- especially the scrollbar ones.


setchy (changes requested):
Thanks @bmulholland 

We spotted a few visual regressions so adding a "Request Changes" review for future us to remember 🎗️

assets/css/main.css
bmulholland, line 1: Can kill this file now :)

PR #1787, opened 2025-01-29
refactor: remove legacy account model and migration util

setchy:
We introduced the new account model in [v5.7.0 (June 5, 2024)](https://github.com/gitify-app/gitify/releases/tag/v5.7.0).

With v6 being a major release this may be an opportune time to remove the deprecated account model and supporting migration service


afonsojramos:
Seems like enough time has passed 👍

PR #1800, opened 2025-02-01
feat: encrypt tokens using safe storage api

setchy:
Use [Electron's Safe Storage API](https://www.electronjs.org/docs/latest/api/safe-storage) to encrypt and decrypt tokens.

Still supports backwards compatibility.


setchy:
I ported ~95% of this from https://github.com/setchy/atlassify

The 5% newness is handling both unecrypted tokens and encrypted tokens (Atlassify launched with encryption from its first GA release)

bmulholland (commented):
Nice touch, thanks for considering security

PR #1840, opened 2025-02-14
test: migrate from `jest` to `vitest`

setchy:


bmulholland:
FWIW, LLMs are VERY good at conversions like this

PR #1852, opened 2025-02-20
fix: change filter logic to avoid early returning

### LINKED ISSUE #1848: Not filtering by excluded handle (GHE)

### 📝 Description

I have a handle exclusion specified in filters yet am still receiving notifications for PRs from that user.

PS thanks for Gitify, love it!

### 🪜 Steps To Reproduce

1. Using GitHub Enterprise account
2. Ensure detailed notifications is on
3. Add a handle to exclude
4. Create a PR or other notification with that user
5. Notification appears in list
### 🪵 Log Excerpts

_No response_

### Gitify Version

6.1.0

### Operating System

macOS

### GitHub Account

GitHub Enterprise Server

### 📸 Screenshots

Image

Image

Image

setchy:
resolves #1848


src/renderer/utils/notifications/filters/filter.ts
```diff
@@ -13,33 +13,43 @@ export function filterNotifications(
...
+          settings.filterUserTypes.some((userType) =>
+            filterNotificationByUserType(notification, userType),
+          );
       }
 
       if (hasIncludeHandleFilters(settings)) {
-        return settings.filterIncludeHandles.some((handle) =>
-          filterNotificationByHandle(notification, handle),
-        );
+        passesFilters =
```
grahamj, line 28: You probably want included handles to override the user type check, eg. if you don't check User but have an included handle you'd want to pass.

so: !exclude && (type || include) && reason

setchy: Why not leave the logic as-is and treat all selected filters as an `AND` 🤔

setchy: It does mean as you pointed out (thank you for doing so), there is some interplay between `User Type` and the `Handle Filters` - but that seems OK to me.

@afonsojramos @bmulholland - interested to hear your thoughts 👂

grahamj: np, whatever you think :)

afonsojramos: I think it looks good as is. It can probably be improved with a little modularity, but it is fine as it is.

afonsojramos (approved):

PR #1868, opened 2025-02-27
refactor: layout components

setchy:
Refactor some layout components

And importantly, fix our vertical scrollbar regression on Filters, caused by the primer CounterLabel component, which only occurred in certain states.  tldr; that took hours to triage and find. 😅

### DIFF
--- a/src/renderer/components/fields/Checkbox.tsx
+++ b/src/renderer/components/fields/Checkbox.tsx
@@ -1,8 +1,8 @@
+import { Stack } from '@primer/react';
 import type { FC, ReactNode } from 'react';
 
-import { CounterLabel, Stack } from '@primer/react';
-
 import { cn } from '../../utils/cn';
+import { CustomCounter } from '../primitives/CustomCounter';
 import { Tooltip } from './Tooltip';
 
 export interface ICheckbox {
@@ -20,7 +20,7 @@ export const Checkbox: FC<ICheckbox> = ({
   visible = true,
   ...props
 }: ICheckbox) => {
-  const counter = props.counter > 0 ? props.counter : undefined;
+  const counter = props?.counter === 0 ? '0' : props.counter;
 
   return (
     visible && (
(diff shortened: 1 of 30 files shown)


src/renderer/components/fields/Checkbox.tsx
```diff
@@ -20,7 +20,7 @@ export const Checkbox: FC<ICheckbox> = ({
   visible = true,
   ...props
 }: ICheckbox) => {
-  const counter = props.counter > 0 ? props.counter : undefined;
+  const counter = props?.counter === 0 ? '0' : props.counter;
```
setchy, line 23: When `0` is left as a number, it will not be rendered within the counter label correctly

src/renderer/components/primitives/CustomCounter.tsx
```diff
@@ -0,0 +1,38 @@
...
+  value: string | number;
+  /** The visual style of the counter */
+  scheme?: CounterScheme;
+}
+
+/**
+ * CustomCounter is a component that displays a small count or numeric indicator,
+ * similar to CounterLabel from @primer/react but with customizable styling.
+ *
+ * Created due to odd behavior with CounterLabel:
```
setchy, line 20: literally hours to figure this out 😭

src/renderer/components/primitives/CustomCounter.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * - would not render '0' within a counter.
+ */
+export const CustomCounter: FC<ICustomCounter> = ({
+  value,
+  scheme = 'secondary',
+}) => {
+  const baseStyles =
+    'px-1.5 py-0.75 rounded-full text-[10px] font-medium leading-none min-w-[16px]';
+
+  // TODO - use design tokens for these colors
```
setchy, line 31: I still need to do this - in the morning...

src/renderer/components/fields/Checkbox.tsx
bmulholland: Sounds painful to figure out.

Worth leaving this as a code comment?

PR #1917, opened 2025-03-25
refactor: electron-builder configuration

setchy:
Extract electron-builder configuration from package.json to dedicated js file and add typechecking.


scripts/afterPack.js
```diff
@@ -2,15 +2,19 @@ const path = require('node:path');
...
 const { AfterPackContext } = require('electron-builder');
 
-const packageJson = require('../package.json');
-const electronLanguages = packageJson.build.electronLanguages;
+const builderConfig = require('../config/electron-builder');
+const electronLanguages = builderConfig.electronLanguages;
+
+function logAfterPackProgress(msg) {
+  // biome-ignore lint/suspicious/noConsoleLog: log notarizing progress
+  console.log(`  • [afterSign]: ${msg}`);
```
bmulholland, line 10: This previously said afterPack, and the function name matches that, but the log now says afterSign. Is that intentional?

bmulholland:
Did my best but hard to review. Would have been easier to review if the code moving was a separate commit from the refactors, so I could see the deltas.

scripts/afterPack.js
setchy: Ah great catch.  Copy/paste error

PR #1953, opened 2025-04-09
fix: context menu visibility regression in electron

setchy:
Fix regression introduced in electron post 34.4.1

See https://github.com/electron/electron/issues/46593


setchy:
closing as fixed in https://github.com/electron/electron/releases/tag/v35.1.5

PR #2023, opened 2025-05-26
refactor: volume up and down icons

SiriusCrain:
Adapt volume icon for dark theme, not ideal but better than it is now
Screenshot From 2025-05-26 16-30-15
Screenshot From 2025-05-26 16-00-50


setchy (approved):
Thank you @SiriusCrain 👏

PR #2026, opened 2025-05-26
fix: fixed startup for Gnome DE

SiriusCrain:


src/main/main.ts
```diff
@@ -11,6 +11,8 @@ import { TrayIcons } from './icons';
 import MenuBuilder from './menu';
 import Updater from './updater';
 
+if (isLinux) app.commandLine?.appendSwitch('--gtk-version', '3');
```
afonsojramos, line 14:
```suggestion
# https://github.com/electron/electron/issues/46538
if (isLinux) app.commandLine?.appendSwitch('--gtk-version', '3');
```
Added reference to issue. @SiriusCrain other PRs did not have the double dash before gtk-version. Can you confirm that it works without them as well?

PR #2040, opened 2025-06-01
chore: primer react 37

SiriusCrain:


SiriusCrain:
@setchy here is PR

setchy:
@SiriusCrain - does the test suite run for you?

SiriusCrain:
> @SiriusCrain - does the test suite run for you?

I am AFK, will check when can

setchy:
I've also got a handful of other tailwind padding/margin changes locally to contribute later...

SiriusCrain:
> I've also got a handful of other tailwind padding/margin changes locally to contribute later...

 I think you can add them or here or after merge, how will be ok for you

SiriusCrain:
> I've also got a handful of other tailwind padding/margin changes locally to contribute later...

Problem with tests because jest cannot properly handle ESM syntax on which primer rely a lot, so I think it will be easier to migrate from jest first

afonsojramos:
> Problem with tests because jest cannot properly handle ESM syntax on which primer rely a lot, so I think it will be easier to migrate from jest first

We've tried in the past and it was not super easy as we had a lot of the mocking was not working. But I can't completely recall.

SiriusCrain:
> > Problem with tests because jest cannot properly handle ESM syntax on which primer rely a lot, so I think it will be easier to migrate from jest first
> 
> We've tried in the past and it was not super easy as we had a lot of the mocking was not working. But I can't completely recall.

I get it, I am afraid that we cannot upgrade right now because of jest

PR #2135, opened 2025-08-13
fix(discussions): find closest discussion comment or reply by timestamp

setchy:
**Problem**
Earlier today I noticed several discussion notifications which were showing an incorrect actor profile image, and upon notification interaction deep-linking to the wrong discussion item (latest comment instead of most recent reply on an older comment thread).

**Solution**
Updating our logic to find the latest discussion comment or reply to instead find the closest/nearest in timestamp. 

In order to increase the accuracy, we have to fetch a wider selection of comments or replies before filtering down


setchy:
still need to update unit tests, but raising a draft pr for feedback from @afonsojramos @bmulholland

Issue #1997, opened 2025-05-08
Mitigate API hammering with high notification counts


grahamj:
### 📝 Provide a description of the new feature

Hi, you might remember me from such feature suggestions as https://github.com/gitify-app/gitify/issues/1857

Long story short, today we figured out that my use of Gitify has been bringing our GHE to its knees 🐼 For a 1-minute period I see Gitify requesting many pages of notifications and then close to 300 each of pulls and reviews. Root cause is I don't clear out old merged PR notifications because I use Gitify which helpfully filters them.

Now I realize this is what it's supposed to do and it's my bad for not cleaning up my notifications. But in my defence:
- The vast majority of those 300 PRs are merged
- Thanks to your gracious addition of merged filtering I never see them in Gitify
- The main reason I use Gitify is to avoid GH notifications because it doesn't do that filtering

I don't think the notifications API provides PR status (making one wonder how it's displayed on the notifications webpage) so avoiding those requests for merged PRs would seem to be impossible. But a couple of other possibilities that come to mind are:

- Feature that automatically marks filtered notifications as read
- Option to limit considered notifications by number or time
- Option to rate limit API calls to at least smooth out the spikes

So this is more of a discussion starter than request per se. Thanks for Gitify!

setchy:
A great discussion to have - thanks for starting it @grahamj.

To confirm, I assume this option doesn't meet your needs / workflow?

<img width="497" alt="Image" src="https://github.com/user-attachments/assets/8649cd66-4633-4abc-9750-3ecb708e2e78" />

setchy:
> Even if /pulls is necessary maybe don't grab /reviews for filtered PRs

This is a good suggestion, though could be tricky to implement.  At the moment the filtering occurs as a post processing step since some of the filters are actually derived values.

setchy:
API call volumes reduces via #2156 and #2158.  Happy to reopen this issue if there are further opportunities

setchy:
@grahamj - appreciate your feedback re [v6.6.0](https://github.com/gitify-app/gitify/releases/tag/v6.6.0) and if that helps your situation

PR #2156, opened 2025-08-16
feat(filters): split into two stages (base, detailed) to reduce wasted api calls

setchy:
Partially addresses #1997 

Splits the notification filtering into a two-stage process
* Base notification filters (type and reason)
* Detailed notification filters (user type, handles, state)


setchy:
@grahamj - this is part one of reducing api calls - hope you find it useful.

I'll follow-up with a separate to avoid fetching reviews etc if notifications have been filtered by their detailed state

PR #2174, opened 2025-08-26
feat: added the organization filter

vallieres:
Adding an Organization filter (just like the Handle filter)

### DIFF
--- a/src/renderer/components/filters/OrganizationFilter.tsx
+++ b/src/renderer/components/filters/OrganizationFilter.tsx
@@ -15,7 +15,6 @@ import {
 } from '../../utils/notifications/filters/organizations';
 import { Tooltip } from '../fields/Tooltip';
 import { Title } from '../primitives/Title';
-import { RequiresDetailedNotificationWarning } from './RequiresDetailedNotificationsWarning';
 
 type InputToken = {
   id: number;
@@ -143,7 +142,6 @@ export const OrganizationFilter: FC = () => {
           tooltip={
             <Stack direction="vertical" gap="condensed">
               <Text>Filter notifications by organization.</Text>
-              <RequiresDetailedNotificationWarning />
             </Stack>
           }
         />
--- a/src/renderer/components/filters/OrganizationFilter.tsx
+++ b/src/renderer/components/filters/OrganizationFilter.tsx
@@ -1,0 +1,208 @@
+import { type FC, useContext, useEffect, useState } from 'react';
+
+import {
+  CheckCircleFillIcon,
+  NoEntryFillIcon,
+  OrganizationIcon,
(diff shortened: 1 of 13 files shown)


src/renderer/utils/notifications/filters/organizations.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function hasExcludeOrganizationFilters(settings: SettingsState) {
+  return settings.filterExcludeOrganizations.length > 0;
+}
+
+export function filterNotificationByOrganization(
+  notification: Notification,
+  organizationName: string,
+): boolean {
+  return notification.repository.owner.login === organizationName;
```
setchy: @vallieres - what's your thoughts on making this comparison case-insensitive? 

For example, https://github.com/CycloneDX

src/renderer/utils/notifications/filters/filter.ts
```diff
@@ -48,6 +51,23 @@ export function filterDetailedNotifications(
   return notifications.filter((notification) => {
     let passesFilters = true;
 
+    // Organization filters should always apply, regardless of detailed notifications setting
+    if (hasIncludeOrganizationFilters(settings)) {
```
setchy: we can move these into the `filterBaseNotifications` function since they don't require detailed notifications to be enabled

src/renderer/utils/notifications/filters/organizations.ts
vallieres: Good idea, I will do it :D

src/renderer/components/filters/OrganizationFilter.tsx
```diff
@@ -0,0 +1,208 @@
...
+  return (
+    <fieldset id="filter-organizations">
+      <Stack direction="horizontal" gap="condensed" align="baseline">
+        <Title icon={OrganizationIcon}>Organizations</Title>
+        <Tooltip
+          name="tooltip-filter-organizations"
+          tooltip={
+            <Stack direction="vertical" gap="condensed">
+              <Text>Filter notifications by organization.</Text>
+              <RequiresDetailedNotificationWarning />
```
setchy: We can omit this detailed notifications warning excerpt

vallieres: Done!

src/renderer/utils/notifications/filters/filter.ts
vallieres: Done!

src/renderer/utils/notifications/filters/organizations.ts
vallieres: Done!

PR #2178, opened 2025-08-27
feat: auto updates for all platforms

setchy:
This enhancement drops `update-electron-app` and uses `electron-updater` directly.

This should mean that Linux and Windows users will now received automated (scheduled) update support 🚀


setchy:
Ported from https://github.com/setchy/atlassify/pull/1428

PR #2190, opened 2025-08-27
feat: context bridge and node isolation

### LINKED ISSUE #700: Move off of `nodeIntegration` to `contextIsolation:true`

For security reasons, https://www.electronjs.org/docs/latest/tutorial/context-isolation should be enabled, however, some changes are required to the way we communicate with the main thread. We should apply these changes and enable contextIsolation.

### ➕ Additional Information

https://stackoverflow.com/questions/71111106/nodeintegration-vs-preload-js-vs-ipc-in-electron

setchy:
Closes #700

Ports security-focused enhancements that have been running successfully for ~6+ months in [setchy/atlassify](https://github.com/setchy/atlassify)


jest.config.ts
```diff
@@ -5,7 +5,6 @@ const config: Config = {
   setupFilesAfterEnv: ['<rootDir>/src/renderer/__helpers__/jest.setup.ts'],
   testEnvironment: 'jsdom',
   collectCoverage: true,
-  collectCoverageFrom: ['src/**/*', '!**/__snapshots__/**'],
```
afonsojramos: Are we no longer collecting coverage?

src/renderer/components/layout/EmojiSplash.test.tsx
```diff
@@ -1,22 +1,35 @@
 import { render } from '@testing-library/react';
+import { act } from 'react';
```
afonsojramos:
```suggestion
import { act, render } from '@testing-library/react';
```

jest.config.ts
setchy: Added back - good catch

bmulholland: Nice catch.

PR #2227, opened 2025-09-15
feat: working electron-vite config

afonsojramos:
This pull request migrates the Electron app's build system from Webpack to electron-vite, resulting in a significant simplification of the configuration files and build scripts. It removes all Webpack-related configuration and dependencies, introduces new vite-based configuration, and updates file paths and environment variable handling throughout the codebase. Additionally, the CSS and asset handling are modernized to work with Vite and Tailwind v4 (no need for a config file), and logger usage is improved for better compatibility between main and renderer processes.

At the current state however, Primer's variables are being loaded, but somehow not used yet.


src/renderer/App.css
```diff
@@ -4,8 +4,86 @@
 /** GitHub Primer Design System */
 @import "@primer/css/dist/primer.css";
 
-/** Tailwind CSS Configuration */
-@config '../../tailwind.config.ts';
```
bmulholland: It it possible to make the tailwind change separately, to reduce complexity and risk for both vite and tailwind changes?

bmulholland:
I have a working project with electron-vite in use. I'm not super experienced with it, but may be able to help if you run into something tricky.

afonsojramos:
> I have a working project with electron-vite in use. I'm not super experienced with it, but may be able to help if you run into something tricky.

The goal was to simply simplify the project as webpack does present its challenges, so hopefully less tricky things will appear 🙏

afonsojramos:
Closing this in favor of #2479

PR #2266, opened 2025-10-02
fix: icon color when idle

setchy:


src/main/index.ts
```diff
@@ -165,33 +165,26 @@ app.whenReady().then(async () => {
...
+      if (notificationsCount > 0) {
+        setActiveIcon();
+        return;
+      }
 
-  onMainEvent(EVENTS.ICON_IDLE, () => {
-    if (!mb.tray.isDestroyed()) {
       setIdleIcon();
     }
   });
```
0xdevalias, line 185: This is a naive quick skim of this code; but the logic around `setErrorIcon` looks weird. Is that correct?

Might just be the diff making things look more confusing; but currently it sort of looks like the error icon will be used if `notificationsCount < 0`?

0xdevalias: Actually, nevermind. Reading through the rest of it I see that `-1` is used for an error state, so this makes sense.

setchy: Thanks for taking a look. 

That's correct, at the moment a negative number is used to flag that the app is in error state. 

Some further room for improvement

package.json
```diff
@@ -125,7 +125,7 @@
     "webpack-cli": "6.0.1",
     "webpack-merge": "6.0.1"
   },
-  "packageManager": "pnpm@10.18.0",
+  "packageManager": "pnpm@10.17.0",
```
bmulholland, line 128: Was the downgrade intentional?

setchy: Yep, timing issue

Issue #2284, opened 2025-10-09
Make fetch interval configurable


alexey-anufriev:
### 📝 Provide a description of the new feature

Currently, Gitify fetches notifications from all configured repositories at a fixed interval - 1 minute.
When using Gitify with a large number of repositories (I have 40+), this results in a noticeable network traffic.

As I see, there’s no UI setting to adjust this interval, it’s hardcoded in `FETCH_NOTIFICATIONS_INTERVAL_MS`.

**Why does it even matter for me?**

Generated traffic:

- causes load on CPU/network
- may hit GitHub’s rate limits
- unnecessary battery drain on my mac

I would prefer a longer polling interval (e.g., every 5–15 minutes), yet someone may want even shorter intervals for near-real-time updates.

**Proposed Solution**

Make the fetch interval configurable in UI, e.g., “Check for new notifications every: [1m (default) / 5m / 10m / custom]”). Custom will be defined in seconds, lower limit could be 15 seconds, in order to not explode the network.

setchy:
Related https://github.com/gitify-app/gitify/issues/1027

0xdevalias:
> may hit GitHub’s rate limits

I only had a quick skim through the code (assisted by GitHub Copilot GPT-5), but from what I saw, it looks like the app currently doesn't make use of headers such as ETags, `Last-Modified` / `If-Modified-Since`, `X-Poll-Interval`, etc:

- https://docs.github.com/en/rest/activity/notifications
  - > Notifications are optimized for polling with the `Last-Modified` header. If there are no new notifications, you will see a `304 Not Modified` response, leaving your current rate limit untouched. There is an `X-Poll-Interval` header that specifies how often (in seconds) you are allowed to poll. In times of high server load, the time may increase. Please obey the header.

Using these aligns with GitHub's best practices for using their API; and by doing so, if the content isn't modified, it won't count against the primary rate limit:

- https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api#use-conditional-requests-if-appropriate
  - > **Use conditional requests if appropriate**
    >
    > Most endpoints return an `etag` header, and many endpoints return a `last-modified` header. You can use the values of these headers to make conditional `GET` requests. If the response has not changed, you will receive a `304 Not Modified` response. Making a conditional request does not count against your primary rate limit if a `304` response is returned and the request was made while correctly authorized with an `Authorization` header.

[...]

Issue #1437, opened 2024-08-05
Auto Refresh Interval: switch from schedule to inactivity timer


setchy:
### 📝 Provide a description of the new feature

A suggestion was made on a previous Issue that instead of the current behavior which will perform a notification refresh on a 60 seconds schedule, that instead the refresh could be made upon 60 seconds of user inactivity.  

This would prevent scenarios where the newly fetched notifications interferes with the end-users flow (for example, if they were marking notifications as read)

PR #2298, opened 2025-10-15
feat: stabilize notification order during interactions

### LINKED ISSUE #319: Stop re-ordering notifications when I mark one as read

setchy:
Partially closes #319 

Maintains a stable notification order during notification interactions (mark as read, etc) until the next scheduled or manual notification refresh interval.

Separate enhancements around #2284 and #1437 will compliment this enhancement

### DIFF
--- a/src/renderer/utils/notifications/notifications.ts
+++ b/src/renderer/utils/notifications/notifications.ts
@@ -146,6 +146,13 @@ export async function enrichNotification(
   };
 }
 
+/**
+ * Assign an order property to each notification to stabilize how they are displayed
+ * during notification interaction events (mark as read, mark as done, etc.)
+ *
+ * @param notifications
+ * @param settings
+ */
 export function stabilizeNotificationsOrder(
   notifications: AccountNotifications[],
   settings: SettingsState,


src/renderer/components/notifications/AccountNotifications.tsx
```diff
@@ -130,7 +130,7 @@ export const AccountNotifications: FC<IAccountNotifications> = (
         <>
           {props.error && <Oops error={props.error} fullHeight={false} />}
           {!hasNotifications && !props.error && <AllRead fullHeight={false} />}
-          {isGroupByRepository
+          {isGroupByRepository(settings)
             ? Object.values(groupedNotifications).map((repoNotifications) => {
```
0xdevalias, line 134: I think this `Object.values` may be redundant now, as `groupedNotifications` is an array rather than an object after the changes in this PR.

PR #2443, opened 2025-12-12
test(jest): set timezone for stable snapshots

setchy:


src/renderer/__helpers__/jest.setup.ts
```diff
@@ -45,10 +45,6 @@ window.gitify = {
   raiseNativeNotification: jest.fn(),
 };
 
-// Mock OAuth client ID and secret
-process.env.OAUTH_CLIENT_ID = 'FAKE_CLIENT_ID_123';
```
setchy, line 49: As best as I can tell, these are unused / not effective

src/renderer/__helpers__/jest.setup.env.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Jest Global Setup
+ * This runs once before all test suites
+ */
+export default () => {
+  // Sets timezone to UTC for consistent date/time in tests and snapshots
+  process.env.TZ = 'UTC';
```
setchy, line 7: Prevent datetimes changing in snapshots when `pnpm test` is run in different timezones (ie: Austrlaia 🦘)

PR #2480, opened 2025-12-28
refactor: add type-safe useAppContext hook

afonsojramos:
## Summary
- Adds `useAppContext()` hook for type-safe context access
- Replaces manual `useContext(AppContext)` calls throughout codebase

## Benefits
- Type-safe: returns `AppContextState` without casting
- Fail-fast: throws if used outside provider
- Cleaner imports: one hook instead of context + useContext

## Test plan
- [x] All existing tests pass
- [x] Lint passes

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -1,4 +1,3 @@
-import { useContext } from 'react';
 import {
   Navigate,
   Route,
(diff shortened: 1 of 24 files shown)


setchy (approved):
🎉

PR #2479, opened 2025-12-27
feat: migrate from Electron to Tauri v2

afonsojramos:
## Summary

This PR completes a full migration of Gitify from **Electron** to **Tauri v2**, replacing the entire desktop application framework while maintaining feature parity and improving security, performance, and maintainability.

### Why migrate from Electron?

The Electron implementation had several significant drawbacks:

**Security Concerns:**
- OAuth client secrets were embedded in the JavaScript bundle, easily extractable via DevTools
- Tokens were stored in localStorage, accessible to any JavaScript running in the renderer
- The full Node.js runtime exposed a large attack surface
- Webpack bundling made it harder to audit what code was actually shipped

**Performance & Size Issues:**
- Electron apps bundle Chromium (~150MB+ on macOS), resulting in large download sizes
- High memory usage due to running a full browser engine
- Slow startup times, especially on older hardware
- Each Electron app is essentially a full Chrome instance

**Developer Experience:**
- Complex Webpack configuration across 3 targets (main, preload, renderer)
- Jest setup required extensive mocking of Electron APIs
- Babel + TypeScript + Webpack created a fragile build pipeline
- Hot reload was unreliable and slow

**Maintenance Burden:**
- Electron releases frequently with breaking changes
- Security updates required frequent Electron bumps
- The `menubar` dependency was unmaintained
- Code signing and notarization scripts were complex and brittle

### What Tauri v2 provides:

**Security Improvements:**
- OAuth secrets compiled into Rust binary (not extractable via DevTools)
- Tokens stored in OS keyring (macOS Keychain, Windows Credential Manager, Linux Secret Service)
- Capability-based permissions model restricts what the frontend can access
- SSRF protection with hostname validation blocking internal network addresses
- HTTP client timeout prevents hanging connections
- Minimal attack surface - no Node.js, just a WebView

**Performance & Size:**
- ~10x smaller binary size (~15MB vs ~150MB)
- Native system WebView (no bundled Chromium)
- Faster startup and lower memory footprint
- Rust backend is compiled to native code

**Developer Experience:**
- Vite for instant HMR and fast builds
- Vitest for modern, fast testing
- Simpler project structure (no main/preload/renderer split)
- TypeScript-first with better type safety through the bridge API
- Single `tauri-bridge.ts` provides a clean abstraction layer

**Maintenance:**
- Tauri has a stable, well-documented API
- Rust's safety guarantees reduce runtime bugs
- Smaller dependency tree
- Built-in auto-updater with signing

### Architecture

[...]

PR #2482, opened 2025-12-29
refactor: use React useId() hook for filter component IDs

afonsojramos:
## Summary
- Replace hardcoded `id` props in filter components with React's `useId()` hook
- Ensures SSR-safe unique IDs following React best practices
- Extracted from #2479 per review feedback

## Changes
- `ReasonFilter`: `id="filter-reasons"` -> `useId()`
- `StateFilter`: `id="filter-state"` -> `useId()`
- `SubjectTypeFilter`: `id="filter-subject-type"` -> `useId()`
- `UserTypeFilter`: `id="filter-user-types"` -> `useId()`

## Test plan
- [x] All existing tests pass
- [x] Snapshots updated to reflect new ID format

### DIFF
--- a/src/renderer/components/filters/ReasonFilter.tsx
+++ b/src/renderer/components/filters/ReasonFilter.tsx
@@ -1,4 +1,4 @@
-import type { FC } from 'react';
+import { type FC, useId } from 'react';
 
 import { NoteIcon } from '@primer/octicons-react';
 import { Text } from '@primer/react';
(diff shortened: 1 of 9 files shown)


setchy (approved):

afonsojramos:
After opening this PR, I did some thinking, instead of trusting the linter, and I think that to be honest, `useId()` doesn't really make sense here. Since this is client-only, no SSR, and `useId()` was designed to solve hydration mismatch issues, specifically, between server and client.

setchy:
Makes sense, good pickup

PR #2485, opened 2025-12-29
fix(renovate): use changelogUrl instead of deprecated sourceUrl

afonsojramos:
## Summary
Fix deprecated `sourceUrl` field in twemoji rule by replacing it with `changelogUrl`.

## Changes
- Replace `sourceUrl` with `changelogUrl` for twemoji package rule (Renovate deprecated `sourceUrl`)

Thanks @setchy for the feedback on the original PR - I've reduced the scope to just this fix since PR #2486 addressed the notification concerns via the Dependency Dashboard approval workflow.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,9 +9,6 @@
   ],
   "labels": ["dependency"],
   "rangeStrategy": "pin",
-  "prConcurrentLimit": 3,
-  "prHourlyLimit": 2,
-  "minimumReleaseAge": "3 days",
   "packageRules": [
     {
       "matchDepTypes": ["engines"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -9,6 +9,9 @@
   ],
   "labels": ["dependency"],
   "rangeStrategy": "pin",
+  "prConcurrentLimit": 3,
+  "prHourlyLimit": 2,
+  "minimumReleaseAge": "3 days",
```
setchy: This is part of the best-practices preset we extend

renovate.json
```diff
@@ -22,7 +25,128 @@
     {
       "description": "Fetch changelog details for twemoji packages",
       "matchPackageNames": ["@discordapp/twemoji"],
-      "sourceUrl": "https://github.com/jdjdecked/twemoji"
+      "changelogUrl": "https://github.com/jdjdecked/twemoji"
```
setchy, line 25: Good catch - the next major release of renovate would have auto migrate this config

setchy:
I have an alternate approach, which I'll propose as a separate PR, that uses the Renovate Dependency Dashboard approval workflow. This should help some of your concerns.

As for the rate limiting and grouping suggestions, I do not want to go down this path... I much prefer having each update as a separate PR (as-is now), since it's then explicitly referenced in the autogenerated release notes without having to deep dive into each PRs contents. I believe the Approval Dashboard workflow will also help in this regard

setchy:
https://github.com/gitify-app/gitify/pull/2486

Copied the config from renovatebot/.gitthub/default.json - it's the same workflow they use for their project

setchy:
https://github.com/gitify-app/gitify/pull/2486

Copied the config from renovatebot/.gitthub/default.json - it's the same workflow they use for their project

renovate.json
```diff
@@ -22,7 +25,128 @@
...
+      "matchPackageNames": ["@babel/**"],
+      "groupName": "babel"
+    },
+    {
+      "description": "Group GraphQL packages",
+      "matchPackageNames": ["graphql", "@graphql-codegen/**"],
+      "groupName": "graphql"
+    },
+    {
+      "description": "Group remaining non-major npm updates not covered by specific groups",
```
setchy: This could be done by using the preset "group:allNonMajor"

https://docs.renovatebot.com/presets-group/#groupallnonmajor

afonsojramos:
> I have an alternate approach, which I'll propose as a separate PR, that uses the Renovate Dependency Dashboard approval workflow. This should help some of your concerns.
> 
> As for the rate limiting and grouping suggestions, I do not want to go down this path... I much prefer having each update as a separate PR (as-is now), since it's then explicitly referenced in the autogenerated release notes without having to deep dive into each PRs contents. I believe the Approval Dashboard workflow will also help in this regard

@setchy I understand that perspective, but given that only when we release/are about to are these changes shipped, wouldn't it make sense to only merge these kinds of PRs then (excluding majors)?

PR #2488, opened 2025-12-30
feat: add 'Show read notifications' setting

### LINKED ISSUE #708: Feature request: add a "Show read notifications" setting

Thanks to the (coming 🤞) new `Mark notification as done` button (see #500 & #706), I think we could add a new "Show read notifications" setting, to better mimic the default Github's notifications experience.

The workflow would be the following:

**With the setting ON:**
1. You get a new notification, it appears in Gitify
2. You click on this notification, it opens and Github and automatically gets marked as "read"
3. It still appears inside Gitify, with a grey background, and you have the ability to click the new `Mark as done` button to make it disappear

**With the setting OFF:**
1. You get a new notification, it appears in Gitify
2. You click on this notification, it opens and Github and automatically gets marked as "read"
3. It doesn't appear inside Gitify anymore

### ➕ Additional Information

_No response_

afonsojramos:
## Summary
Adds a new setting to show read notifications alongside unread ones. Fixes #708.

## Changes
- Added `showReadNotifications` setting (default: **disabled**)
- When enabled, fetches both read and unread notifications via `all=true` API parameter
- Read notifications display at 50% opacity for visual differentiation
- "Mark as read" button is hidden on already-read notifications
- Tooltip includes rate limiting warning for users with many notifications

## Why "Mark as read" isn't a toggle
GitHub's REST API only supports one-way transitions: `unread → read → done`. There's no endpoint to mark a notification as unread, so we can't implement a toggle. See [GitHub API docs](https://docs.github.com/en/rest/activity/notifications).

## Test plan
- [x] Toggle setting on → read notifications appear with reduced opacity
- [x] Toggle setting off → only unread notifications shown (default behavior)
- [x] "Mark as read" button hidden on read notifications
- [x] Unit tests pass

### DIFF
--- a/src/renderer/components/__snapshots__/Sidebar.test.tsx.snap
+++ b/src/renderer/components/__snapshots__/Sidebar.test.tsx.snap
@@ -235,6 +235,45 @@ exports[`renderer/components/Sidebar.tsx should render itself & its children (lo
               </span>
               <button
                 aria-describedby="_r_ID_"
+                aria-label="Toggle focused mode"
+                class="prc-Button-ButtonBase-9n-Xk prc-Button-IconButton-fyge7"
+                data-component="IconButton"
+                data-loading="false"
+                data-no-visuals="true"
+                data-size="small"
+                data-testid="sidebar-focused-mode"
+                data-variant="invisible"
+                type="button"
+              >
+                <svg
+                  aria-hidden="true"
+                  class="octicon octicon-eye"
+                  display="inline-block"
+                  fill="currentColor"
+                  focusable="false"
+                  height="16"
+                  overflow="visible"
+                  style="vertical-align: text-bottom;"
+                  viewBox="0 0 16 16"
+                  width="16"
+                >
+                  <path
+                    d="M8 2c1.981 0 3.671.992 4.933 2.078 1.27 1.091 2.187 2.345 2.637 3.023a1.62 1.62 0 0 1 0 1.798c-.45.678-1.367 1.932-2.637 3.023C11.67 13.008 9.981 14 8 14c-1.981 0-3.671-.992-4.933-2.078C1.797 10.83.88 9.576.43 8.898a1.62 1.62 0 0 1 0-1.798c.45-.677 1.367-1.931 2.637-3.022C4.33 2.992 6.019 2 8 2ZM1.679 7.932a.12.12 0 0 0 0 .136c.411.622 1.241 1.75 2.366 2.717C5.176 11.758 6.527 12.5 8 12.5c1.473 0 2.825-.742 3.955-1.715 1.124-.967 1.954-2.096 2.366-2.717a.12.12 0 0 0 0-.136c-.412-.621-1.242-1.75-2.366-2.717C10.824 4.242 9.473 3.5 8 3.5c-1.473 0-2.825.742-3.955 1.715-1.124.967-1.954 2.096-2.366 2.717ZM8 10a2 2 0 1 1-.001-3.999A2 2 0 0 1 8 10Z"
+                  />
+                </svg>
+              </button>
+              <span
+                aria-hidden="true"
+                class="prc-TooltipV2-Tooltip-tLeuB"
+                data-direction="e"
+                id="_r_ID_"
+                popover="auto"
+                role="tooltip"
+              >
+                Participating and watching
+              </span>
+              <button
+                aria-describedby="_r_ID_"
                 aria-label="Filters"
                 class="prc-Button-ButtonBase-9n-Xk prc-Button-IconButton-fyge7"
                 data-component="IconButton"
@@ -573,6 +612,45 @@ exports[`renderer/components/Sidebar.tsx should render itself & its children (lo
               role="tooltip"
             >
               0 unread notifications ↗
+            </span>
+            <button
+              aria-describedby="_r_ID_"
+              aria-label="Toggle focused mode"
+              class="prc-Button-ButtonBase-9n-Xk prc-Button-IconButton-fyge7"
+              data-component="IconButton"
+              data-loading="false"
+              data-no-visuals="true"
+              data-size="small"
+              data-testid="sidebar-focused-mode"
+              data-variant="invisible"
+              type="button"
+            >
+              <svg
+                aria-hidden="true"
+                class="octicon octicon-eye"
+                display="inline-block"
+                fill="currentColor"
+                focusable="false"
+                height="16"
+                overflow="visible"
+                style="vertical-align: text-bottom;"
+                viewBox="0 0 16 16"
+                width="16"
+              >
+                <path
+                  d="M8 2c1.981 0 3.671.992 4.933 2.078 1.27 1.091 2.187 2.345 2.637 3.023a1.62 1.62 0 0 1 0 1.798c-.45.678-1.367 1.932-2.637 3.023C11.67 13.008 9.981 14 8 14c-1.981 0-3.671-.992-4.933-2.078C1.797 10.83.88 9.576.43 8.898a1.62 1.62 0 0 1 0-1.798c.45-.677 1.367-1.931 2.637-3.022C4.33 2.992 6.019 2 8 2ZM1.679 7.932a.12.12 0 0 0 0 .136c.411.622 1.241 1.75 2.366 2.717C5.176 11.758 6.527 12.5 8 12.5c1.473 0 2.825-.742 3.955-1.715 1.124-.967 1.954-2.096 2.366-2.717a.12.12 0 0 0 0-.136c-.412-.621-1.242-1.75-2.366-2.717C10.824 4.242 9.473 3.5 8 3.5c-1.473 0-2.825.742-3.955 1.715-1.124.967-1.954 2.096-2.366 2.717ZM8 10a2 2 0 1 1-.001-3.999A2 2 0 0 1 8 10Z"
+                />
+              </svg>
+            </button>
+            <span
+              aria-hidden="true"
+              class="prc-TooltipV2-Tooltip-tLeuB"
+              data-direction="e"
+              id="_r_ID_"
+              popover="auto"
+              role="tooltip"
+            >
+              Participating and watching
             </span>
             <button
               aria-describedby="_r_ID_"
--- a/src/renderer/components/layout/__snapshots__/AppLayout.test.tsx.snap
+++ b/src/renderer/components/layout/__snapshots__/AppLayout.test.tsx.snap
@@ -149,6 +149,45 @@ exports[`renderer/components/layout/AppLayout.tsx should render itself & its chi
                 </span>
                 <button
                   aria-describedby="_r_ID_"
+                  aria-label="Toggle focused mode"
+                  class="prc-Button-ButtonBase-9n-Xk prc-Button-IconButton-fyge7"
+                  data-component="IconButton"
+                  data-loading="false"
+                  data-no-visuals="true"
+                  data-size="small"
+                  data-testid="sidebar-focused-mode"
+                  data-variant="invisible"
+                  type="button"
+                >
+                  <svg
+                    aria-hidden="true"
+                    class="octicon octicon-eye"
+                    display="inline-block"
+                    fill="currentColor"
+                    focusable="false"
+                    height="16"
+                    overflow="visible"
+                    style="vertical-align: text-bottom;"
+                    viewBox="0 0 16 16"
+                    width="16"
+                  >
+                    <path
+                      d="M8 2c1.981 0 3.671.992 4.933 2.078 1.27 1.091 2.187 2.345 2.637 3.023a1.62 1.62 0 0 1 0 1.798c-.45.678-1.367 1.932-2.637 3.023C11.67 13.008 9.981 14 8 14c-1.981 0-3.671-.992-4.933-2.078C1.797 10.83.88 9.576.43 8.898a1.62 1.62 0 0 1 0-1.798c.45-.677 1.367-1.931 2.637-3.022C4.33 2.992 6.019 2 8 2ZM1.679 7.932a.12.12 0 0 0 0 .136c.411.622 1.241 1.75 2.366 2.717C5.176 11.758 6.527 12.5 8 12.5c1.473 0 2.825-.742 3.955-1.715 1.124-.967 1.954-2.096 2.366-2.717a.12.12 0 0 0 0-.136c-.412-.621-1.242-1.75-2.366-2.717C10.824 4.242 9.473 3.5 8 3.5c-1.473 0-2.825.742-3.955 1.715-1.124.967-1.954 2.096-2.366 2.717ZM8 10a2 2 0 1 1-.001-3.999A2 2 0 0 1 8 10Z"
+                    />
+                  </svg>
+                </button>
+                <span
+                  aria-hidden="true"
+                  class="prc-TooltipV2-Tooltip-tLeuB"
+                  data-direction="e"
+                  id="_r_ID_"
+                  popover="auto"
+                  role="tooltip"
+                >
+                  Participating and watching
+                </span>
+                <button
+                  aria-describedby="_r_ID_"
                   aria-label="Filters"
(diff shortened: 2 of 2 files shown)


setchy:
I've had this on my half-implemented list for a while, with a few raised and then closed PR attemps... 

I think we need to wait until the merged/batched api is finished before merging to avoid rate limits. 

Additionally, in the previous PRs I raised for this feature, I had changes to the HoverGroup buttons to hide the "mark as read" option on read NotificationRows

setchy:
We should also change the look and feel for read NotificationRows - ie: opacity difference, plus as Read State as a Filter Section option for Read and Unread. 

Unfortunately the REST API doesn't indicate a difference between read and done notification types

setchy:
> Unfortunately the REST API doesn't indicate a difference between read and done notification types

To add some further thoughts for discussion, when I last looked into this feature ~2 months or so back that I was 80% convinced to close #708 as `not planned` since I couldn't see it being that useful alone unless we can find a way to differentiate `read + done` vs `read and undone`, or essentially the `not done inbox`.  I timeboxed this but wasn't able to find any attributes within the public REST or GraphQL API endpoints that could help us here...

afonsojramos:
@setchy, alright, I think I've landed on a good implementation here. 

1. Default is now false (opt-in, preserves existing behavior)
2. "Mark as read" button is hidden on already-read notifications
3. Read notifications already render at 50% opacity
4. Added all=true API param when setting is enabled to actually fetch read notifications
5. Added rate limiting warning in the tooltip

On the read vs done question:
Done notifications are DELETEd from the API entirely - they never appear in responses. So, the unread boolean cleanly distinguishes read from unread. The concern about differentiating "read + done" vs "read + undone" doesn't apply since done items are gone.

On rate limiting:
The tooltip warns users, and it's opt-in. Happy to consider deferring to when the batched API is ready if you think that's better.

What do you think?

src/renderer/components/settings/NotificationSettings.tsx
```diff
@@ -328,6 +328,31 @@ export const NotificationSettings: FC = () => {
           }
         />
 
+        <Checkbox
+          checked={settings.showReadNotifications}
+          label="Show read notifications"
```
setchy: "Fetch read notifications" to be consistent with "Fetch only participating" setting?

src/renderer/utils/notifications/filters/filter.ts
```diff
@@ -24,6 +24,11 @@ export function filterBaseNotifications(
   return notifications.filter((notification) => {
     let passesFilters = true;
 
+    // Filter out read notifications if showReadNotifications is disabled
+    if (!settings.showReadNotifications && !notification.unread) {
```
setchy: This is redundant and can be removed as `settings.showReadNotifications` controls the API query parameters

src/renderer/components/notifications/NotificationRow.tsx
```diff
@@ -148,6 +148,7 @@ export const NotificationRow: FC<NotificationRowProps> = ({
 
           <HoverButton
             action={actionMarkAsRead}
+            enabled={!isNotificationRead}
```
setchy, line 145: The position of the 3 hover group buttons should be consistent between read and unread notification rows.  This currently hides the "mark as read" button and shifts the position of the "mark as done" button

setchy:
In addition to the above comments, we should also:
* Change the Sidebar text when `settings.showReadNotifications` is enabled to read "XX notifications" instead of "XX unread notifications"
* Add a new filter section for "Read" and "Unread"

src/renderer/utils/notifications/filters/filter.ts
setchy: In it's place, we should add a new Filter Section for "Read" and "Unread" which runs on the base notifications

src/renderer/components/notifications/NotificationRow.tsx
setchy: It may be easiest to switch the order of "mark as read" and "mark as done" hover buttons to achieve the same consistent result and avoid having a "gap"

src/renderer/components/settings/NotificationSettings.tsx
afonsojramos: Actually, I do like the fetch word for the read notifications, but maybe for only participating, a better word would fit, one that implies the filtering. Maybe something like "Limit to only participating"?

src/renderer/components/notifications/NotificationRow.tsx
afonsojramos: Moved the mark as done to the far right. I think it works best that way.

src/renderer/utils/notifications/filters/filter.ts
afonsojramos: I thought about this, but I think it might be better as a follow-up. The filter would only be useful when `showReadNotifications` is enabled (otherwise, read notifications aren't even fetched), which could be confusing. The current implementation is self-contained - the setting controls what's fetched, and opacity shows the difference.

src/renderer/components/notifications/NotificationRow.tsx
setchy: After further testing, i noticed two more items
1 - the latest order now makes the notification row hover group inconsistent with the repository row hover group.  
2 - if all notifications are "read", then we should probably hide the repository-level "mark as read" option

src/renderer/utils/api/client.ts
```diff
@@ -67,6 +67,10 @@ export function listNotificationsForAuthenticatedUser(
   url.pathname += 'notifications';
   url.searchParams.append('participating', String(settings.participating));
 
+  if (settings.showReadNotifications) {
```
setchy:
```suggestion 
  url.searchParams.append('all', String(settings.fetchReadNotifications));

```

To match how we've been building the search params for `participating`

setchy (changes requested):

src/renderer/context/defaults.ts
```diff
@@ -36,6 +36,7 @@ const defaultNotificationSettings: NotificationSettingsState = {
   showPills: true,
   showNumber: true,
   participating: false,
+  showReadNotifications: false,
```
setchy:
```suggestion
  fetchReadNotifications: false,
```

src/renderer/types.ts
```diff
@@ -109,6 +109,7 @@ export interface NotificationSettingsState {
   showPills: boolean;
   showNumber: boolean;
   participating: boolean;
+  showReadNotifications: boolean;
```
setchy:
```suggestion
  fetchReadNotifications: boolean,
```

src/renderer/components/settings/NotificationSettings.test.tsx
```diff
@@ -271,6 +271,22 @@ describe('renderer/components/settings/NotificationSettings.tsx', () => {
     );
   });
 
+  it('should toggle the showReadNotifications checkbox', async () => {
```
setchy:
```suggestion
  it('should toggle the fetchReadNotifications checkbox', async () => {
```

src/renderer/components/settings/NotificationSettings.test.tsx
```diff
@@ -271,6 +271,22 @@ describe('renderer/components/settings/NotificationSettings.tsx', () => {
...
+    await act(async () => {
+      renderWithAppContext(<NotificationSettings />, {
+        updateSetting: updateSettingMock,
+      });
+    });
+
+    await userEvent.click(screen.getByTestId('checkbox-showReadNotifications'));
```
setchy:
```suggestion
    await userEvent.click(screen.getByTestId('checkbox-fetchReadNotifications'));
```

src/renderer/components/settings/NotificationSettings.test.tsx
```diff
@@ -271,6 +271,22 @@ describe('renderer/components/settings/NotificationSettings.tsx', () => {
...
+
+    expect(updateSettingMock).toHaveBeenCalledTimes(1);
+    expect(updateSettingMock).toHaveBeenCalledWith(
+      'showReadNotifications',
```
setchy:
```suggestion
      'fetchReadNotifications',
```

src/renderer/components/Sidebar.tsx
```diff
@@ -91,7 +91,7 @@ export const Sidebar: FC = () => {
         <IconButton
           aria-label="Notifications"
           data-testid="sidebar-notifications"
-          description={`${unreadNotificationCount} unread notifications ↗`}
+          description={`${unreadNotificationCount} ${settings.showReadNotifications ? 'notifications' : 'unread notifications'} ↗`}
```
setchy:
```suggestion
          description={`${notificationCount} ${settings.fetchReadNotifications ? 'notifications' : 'unread notifications'} ↗`}
```

src/renderer/components/settings/NotificationSettings.tsx
```diff
@@ -328,6 +328,31 @@ export const NotificationSettings: FC = () => {
           }
         />
 
+        <Checkbox
+          checked={settings.showReadNotifications}
```
setchy:
```suggestion
          checked={settings.fetchReadNotifications}
```

src/renderer/components/settings/NotificationSettings.tsx
```diff
@@ -328,6 +328,31 @@ export const NotificationSettings: FC = () => {
...
+          label="Fetch read notifications"
+          name="showReadNotifications"
```
setchy:
```suggestion
          name="fetchReadNotifications"
```

src/renderer/components/settings/NotificationSettings.tsx
```diff
@@ -328,6 +328,31 @@ export const NotificationSettings: FC = () => {
...
+          onChange={(evt) =>
+            updateSetting('showReadNotifications', evt.target.checked)
```
setchy:
```suggestion
            updateSetting('fetchReadNotifications', evt.target.checked)
```

src/renderer/components/Sidebar.tsx
setchy: This should also be using `notificationCount` from useAppContext.

Similarly in context/App.tsx L294 should be 
```
    const trayCount = status === 'error' ? -1 : notificationCount;
```

setchy (changes requested):

setchy:
> > Unfortunately the REST API doesn't indicate a difference between read and done notification types
> 
> To add some further thoughts for discussion, when I last looked into this feature ~2 months or so back that I was 80% convinced to close #708 as `not planned` since I couldn't see it being that useful alone unless we can find a way to differentiate `read + done` vs `read and undone`, or essentially the `not done inbox`. I timeboxed this but wasn't able to find any attributes within the public REST or GraphQL API endpoints that could help us here...

The attached grab better explains what i poorly trying to explain 😅 - two observed behaviors
1 - When marking a read notification as done, the notification does not get removed it from the list notifications response.  The https://github.com/notifications correctly differentiates between `read` and `done`, but the list notifications API response doesn't differentiate the two.
2 - when marking a notification as read and `settings.fetchReadNotifications` is enabled, then the notification row should be set as Opacity.READ and the exit transition animation prevented

https://github.com/user-attachments/assets/563575d4-d51b-4c21-b934-0ff1ad7bb743

afonsojramos:
> 1 - When marking a read notification as done, the notification does not get removed it from the list notifications response. The [github.com/notifications](https://github.com/notifications) correctly differentiates between read and done, but the list notifications API response doesn't differentiate the two.
2 - when marking a notification as read and settings.fetchReadNotifications is enabled, then the notification row should be set as Opacity.READ and the exit transition animation prevented

Issue number 2 is easily fixable, but issue number one is a doozy... Here are the options I can think of:
1. Accept the limitation - document it for users
2. Also suppress done animation when fetchReadNotifications=true (prevents confusing "disappear then reappear")
3. Track done notifications locally (more complex, scope creep) but likely the best one, long term... What do you think @setchy ?

setchy:
> Issue number 2 is easily fixable, but issue number one is a doozy... Here are the options I can think of:
> 
>     1. Accept the limitation - document it for users
> 
>     2. Also suppress done animation when fetchReadNotifications=true (prevents confusing "disappear then reappear")
> 
>     3. Track done notifications locally (more complex, scope creep) but likely the best one, long term... What do you think @setchy ?

Agree, that one is a huge doozy 🥴 

I think although we _could_  implement `option (3) - local done state tracking`, it would pretty easily drift from that of the official sources, which would deliver an inconsistent user experience.
* GitHub UI `done` list - https://github.com/notifications?query=is%3Adone
* GitHub iOS and Android mobile apps.

My suggestion paths forward would be roughly along these lines

**Suggested Path 1 - Accept the limitation and document caveats**
- Rename the setting key and title from `Fetch read notifications` to something like:
  - `Fetch read & done notifications`.
  - `Fetch all notifications`, but this would require us to rephrase the existing option which unpaginates the full inbox list
- Document the limitation for users via the setting tooltip, explaining that when this new setting is enabled, `read`, `unread` and `done` notifications will be shown and that we aren't able to discern the difference between `read` and `done` at this stage
- For read notifications, 
  - Prevent all exit transition animations (open notification, mark as x, etc)
  - Also hide the `Mark as done` option?  I feel this is a less confusing UX then always showing it without any feedback on `done` vs `undone`


**Suggested Path 2 - Mark as not planned**
- Due to the inability to differentiate `read` vs `done` via the known public API endpoints (ie: https://api.github.com/notifications?all=true), leave this feature in the backburner and/or flag as not planned to be implemented for the challenges described above

setchy:
I've also updated our https://github.com/gitify-app/gitify/issues/890 to document the `?all=true` read vs done challenges.

src/renderer/hooks/useNotifications.ts
```diff
@@ -222,7 +222,12 @@ export const useNotifications = (): NotificationsState => {
           notification.account.token,
         );
 
-        if (state.settings.markAsDoneOnUnsubscribe) {
+        // Fall back to mark as read when fetchReadNotifications is enabled
+        // since marking as done won't work correctly (API limitation)
+        if (
```
setchy, line 245: in addition, i think you'll need to update the `markNotificationsAsRead` function above.  during testing it's still calling the `removeNotificationsForAccount`

```
const updatedNotifications = removeNotificationsForAccount(
      readNotifications[0].account,
      state.settings,
      readNotifications,
      notifications,
    );
```

src/renderer/components/notifications/NotificationRow.tsx
```diff
@@ -137,21 +147,25 @@ export const NotificationRow: FC<NotificationRowProps> = ({
...
             icon={ReadIcon}
             label="Mark as read"
             testid="notification-mark-as-read"
           />
 
+          <HoverButton
+            action={actionMarkAsDone}
+            enabled={
+              isMarkAsDoneFeatureSupported(notification.account) &&
+              !settings.fetchReadNotifications
```
setchy, line 155:
```suggestion
              !isNotificationRead
```

src/renderer/components/notifications/RepositoryNotifications.tsx
```diff
@@ -91,23 +95,25 @@ export const RepositoryNotifications: FC<RepositoryNotificationsProps> = ({
...
               icon={ReadIcon}
               label="Mark repository as read"
               testid="repository-mark-as-read"
             />
 
+            <HoverButton
+              action={actionMarkAsDone}
+              enabled={
+                isMarkAsDoneFeatureSupported(repoNotifications[0].account) &&
+                !settings.fetchReadNotifications
```
setchy, line 110:
```suggestion
                !areAllRepoNotificationsRead}
```

src/renderer/hooks/useNotifications.ts
setchy: Actually, thinking about this further, I believe we can add additional condition can be added to https://github.com/gitify-app/gitify/blob/main/src/renderer/utils/notifications/remove.ts#L32-L37 which is where we already update `unread: false` when `settings.delayNotificationState` is enabled.

That would keep `useNotifications.ts` a little simpler.

PR #2490, opened 2025-12-31
fix: apply high contrast colors when system theme is selected

### LINKED ISSUE #2489: fix: high contrast setting ignored when system theme is selected
## Description

When the **System** theme is selected and **Increase contrast** is enabled, the high contrast color scheme is not applied. The app falls back to the default `light`/`dark` schemes instead of `light_high_contrast`/`dark_high_contrast`.

## Steps to Reproduce

1. Open Settings → Appearance
2. Select **System** theme
3. Enable **Increase contrast**
4. Observe that the UI does not show high contrast colors

## Expected Behavior

When System theme is selected with high contrast enabled:
- In light mode: `light_high_contrast` scheme should be applied
- In dark mode: `dark_high_contrast` scheme should be applied

## Actual Behavior

The app uses default `light` and `dark` schemes regardless of the high contrast setting.

## Root Cause

In `src/renderer/utils/theme.ts`, the `mapThemeModeToColorScheme()` function returns `null` for `Theme.SYSTEM` unconditionally (line 53), bypassing the high contrast logic entirely.

## Proposed Fix

Add a new helper function that handles the System theme case by returning both day and night schemes with proper high contrast support.

afonsojramos:
## Summary

Fixes #2489

When "System" theme is selected with "Increase contrast" enabled, the high contrast color schemes were not being applied. The app would fall back to default `light`/`dark` schemes instead of `light_high_contrast`/`dark_high_contrast`.

This adds high contrast fallback constants and uses them when the system theme returns `null` for the color scheme.

## Test plan

- Select "System" theme in Settings → Appearance
- Enable "Increase contrast"
- Verify high contrast colors are applied in both light and dark system modes

### DIFF
--- a/src/renderer/context/App.tsx
+++ b/src/renderer/context/App.tsx
@@ -55,7 +55,9 @@ import {
 import { clearState, loadState, saveState } from '../utils/storage';
 import {
   DEFAULT_DAY_COLOR_SCHEME,
+  DEFAULT_DAY_HIGH_CONTRAST_COLOR_SCHEME,
   DEFAULT_NIGHT_COLOR_SCHEME,
+  DEFAULT_NIGHT_HIGH_CONTRAST_COLOR_SCHEME,
   mapThemeModeToColorMode,
   mapThemeModeToColorScheme,
 } from '../utils/theme';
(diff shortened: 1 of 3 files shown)


setchy (approved):

afonsojramos:
@setchy found this one out testing the tauri branch, but found out that it actually already existed 🙈

PR #2502, opened 2026-01-05
feat: format display attributes on GitifyNotification

setchy:
Add a new `display` section to our `GitifyNotification` type and a final formatting stage.

Notification stages are now: `fetch` → `transform` → `base filter` → `enrich` → `detail filter` → `format` → `ordering/stabilizing`

Moved across several of our formatting calls to be part of the `formatNotification` function [note: there are more, these will follow in follow up PRs].  

As part of this refactor, I have also simplified the handler interface to focus on parts that vary between notification types (previously it was also performing some base formatting duties)

### DIFF
--- a/src/renderer/utils/notifications/formatters.ts
+++ b/src/renderer/utils/notifications/formatters.ts
@@ -1,6 +1,14 @@
 import type { GitifyNotification } from '../../types';
 import { createNotificationHandler } from './handlers';
 
+/**
+ * Populates the display property on a notification with formatted,
+ * UI-ready values. Must be called after enrichment has completed,
+ * as formatting depends on enriched subject data (state, number, etc.).
+ *
+ * @param notification - The enriched notification to format
+ * @returns The notification with populated display property
+ */
 export function formatNotification(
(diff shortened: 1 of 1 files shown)


src/renderer/utils/notifications/formatters.ts
```diff
@@ -0,0 +1,84 @@
+import type { GitifyNotification } from '../../types';
+import { createNotificationHandler } from './handlers';
+
+export function formatNotification(
```
afonsojramos, line 12:
```suggestion
/**
 * Populates the display property on a notification with formatted,
 * UI-ready values. Must be called after enrichment has completed,
 * as formatting depends on enriched subject data (state, number, etc.).
 *
 * @param notification - The enriched notification to format
 * @returns The notification with populated display property
 */
export function formatNotification(
```

Shall we leave some JSDoc so that it is clear when this is meant to be called?

setchy: Great suggestion. Will add some shortly

PR #2515, opened 2026-01-11
feat(shortcuts): add key bindings and hints

setchy:
Add hotkey bindings for:
* Home = `H`
* My Notifications = `N`
* Focused = `W`
* Filters = `F`
* My Issues = `I`
* My Pull Requests = `P`
* Refresh = `R`
* Settings = `S`
* Quit = `Q`
* Accounts = `A`

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -41,8 +41,8 @@ export const App = () => {
     <ThemeProvider>
       <BaseStyles>
         <AppProvider>
-          <GlobalShortcuts />
           <Router>
+            <GlobalShortcuts />
             <AppLayout>
               <Routes>
                 <Route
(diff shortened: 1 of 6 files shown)


setchy:
@afonsojramos - what's your thoughts on having these hotkeys/shortcuts registered to a single key?  or should we pair it with a combination meta key like `⌘`?

I think it works well without a meta/modifier key

setchy:
> @afonsojramos - what's your thoughts on having these hotkeys/shortcuts registered to a single key? or should we pair it with a combination meta key like `⌘`?
> 
> I think it works well without a meta/modifier key

I've thought about it more and think it's best without modifiers, since those commands are likely to overlap with the chrome shortcuts (like cmd + r)

PR #2573, opened 2026-01-26
fix: inbox flicker during loading

setchy:
Fix a long standing minor annoyance whereby the inbox would "flicker" during the cycle from `loading` and `success` | `error`

### DIFF
--- a/src/renderer/hooks/useNotifications.ts
+++ b/src/renderer/hooks/useNotifications.ts
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useState } from 'react';
+import { useCallback, useMemo, useRef, useState } from 'react';
 
 import type {
   Account,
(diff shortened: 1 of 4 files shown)


src/renderer/hooks/useNotifications.ts
```diff
@@ -138,6 +137,7 @@ export const useNotifications = (): NotificationsState => {
         }
 
         setStatus('success');
+        setGlobalError(null);
```
setchy, line 140: Clear global error only if we made it to the end successfully

src/renderer/routes/Notifications.tsx
```diff
@@ -14,28 +14,57 @@ export const NotificationsRoute: FC = () => {
...
+      notifications,
+      status,
+      globalError,
+      hasNotifications,
+    };
+  }
+
+  // Use previous state if loading
+  const displayState =
+    status === 'loading'
```
setchy, line 37: Use previous state while loading

PR #2617, opened 2026-02-16
feat: migrate from `webpack` to `vite`

setchy:
Migrate build from `webpack` to `vite` for HMR, simpler setup and more goodness - https://vite.dev/

Continue to use `electron-builder` for packaging and publishing.

Note: development vs production localStorage is now separated since dev runs via a local vite server 

Ported from https://github.com/setchy/atlassify/pull/2245


afonsojramos (approved):
Love it ❤️‍🔥

PR #2649, opened 2026-03-02
chore: more graceful error on missing GITHUB_TOKEN in codegen

codebytere:
As in title - this makes this a bit more graceful.

### DIFF
--- a/codegen.ts
+++ b/codegen.ts
@@ -3,11 +3,20 @@ import dotenv from 'dotenv';
 
 dotenv.config();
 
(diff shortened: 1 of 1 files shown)


codegen.ts
```diff
@@ -3,11 +3,19 @@ import dotenv from 'dotenv';
 
 dotenv.config();
 
+if (!process.env.GITHUB_TOKEN) {
+  console.warn(
```
setchy, line 8: Let's add a Biome ignore comment and we're good to go

PR #2760, opened 2026-04-01
feat: allow window to be resizable

### LINKED ISSUE #2441: Allow the window to be resized

I'd like to have a taller window so that I can scan more of my notification list without having to scroll. As far as I can tell, there isn't a way to do this, at least on macOS.

### ➕ Additional Information

_No response_

xingzihai:
Fixes #2441

This PR allows the Gitify window to be resizable:
- Set WindowConfig.resizable to true
- Remove resizable = false in DevTools closed event
- Update test to expect resizable = true

This improves user experience by allowing users to resize the window according to their needs.


setchy (changes requested):
per https://github.com/gitify-app/gitify/issues/2441#issuecomment-3711158260 - we're not sure if this is a feature we want to support at this time

Issue #2309, opened 2025-10-17
Fail to authenticate in hyprland/wayland


ItzTas:
### 🔍 Is there already an issue for your problem?

- [x] I have checked older issues, open and closed

### 📝 Description

It lauches but when I try to validade with token it fails

https://github.com/user-attachments/assets/4e673afa-c048-47c2-b0f1-95bd11f87164

Also when I try to autenticate with github it doesn't work either

https://github.com/user-attachments/assets/5bb97fd9-f675-430a-ac95-02488fb32ef6

### 🪜 Steps To Reproduce

1. Open gitify
2. Try to autenticate via token or Github

### 🪵 Log Excerpts

Gtk-Message: 21:55:25.215: Failed to load module "colorreload-gtk-module"
Gtk-Message: 21:55:25.215: Failed to load module "window-decorations-gtk-module"
21:55:25.282 › [app updater] Starting updater
21:55:25.287 › [app updater] Checking for updates on application launch
21:55:25.287 › APPIMAGE env is not defined, current application is not an AppImage
21:55:25.288 › checkForUpdatesAndNotify called, downloadPromise is null
Error occurred in handler for 'gitify:safe-storage-encrypt': Error: Error while encrypting the text provided to safeStorage.encryptString. Encryption is not available.
    at /usr/lib/gitify/app.asar/build/main.js:2:247228
    at Session.<anonymous> (node:electron/js2c/browser_init:2:107123)
    at Session.emit (node:events:519:28)
21:55:42.831 › [loginWithPersonalAccessToken] Failed to login with PAT Error: Error invoking remote method 'gitify:safe-storage-encrypt': Error: Error while encrypting the text provided to safeStorage.encryptString. Encryption is not available.

### Gitify Version

6.10.0-1

### Operating System

Linux

### GitHub Account

GitHub Cloud

setchy:
Have you tried the FAQ tip titled "How do I fix 'Encryption is not available' errors on Linux?"

https://gitify.io/faq/

ItzTas:
it works now thanks!!!

Issue #1981, opened 2025-04-29
Encryption is not available


TrayserCassa:
### 🔍 Is there already an issue for your problem?

- [x] I have checked older issues, open and closed

### 📝 Description

I recently notice that I didn't get upgrades anymore and was logged out. I tried to login again but it seems my system has some missing dependencies.

I'm not able to login with OAuth or with the token method and in the logs I see the issue with `Encryption is not available`. I guess the storage system can't be encrypted which is an issue by my system but just wanted to ask if there is maybe a missing requirement or at least capture the error properly to show a hint. 

My guess is that this pr is related to it #1800 but not 100% sure. 

My system is ArchLinux 
Node: v22.14.0

### 🪜 Steps To Reproduce

1. Well install arch linux
2. Have a missing dependency for electron
3. Failed login without any warning or error (Besides terminal/Journal)

### 🪵 Log Excerpts

This error is the failed login for the browser session. If you want I can also add the logs for the PAT but I guess it's the same issue.

```
[98103:0429/114254.667532:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 1 times!
[98103:0429/114257.685543:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 2 times!
[98103:0429/114257.688279:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 3 times!
Opening in existing browser session.
11:43:01.305 › [main:second-instance] Second instance was launched.  extracting command to forward
11:43:01.308 › [main:handleUrl] forwarding URL gitify://auth?code=<some-code> to renderer process
11:43:01.308 › [renderer:auth-callback] received authentication callback URL gitify://auth?code=<some-code>
Error occurred in handler for 'gitify:safe-storage-encrypt': Error: Error while encrypting the text provided to safeStorage.encryptString. Encryption is not available.
    at /usr/lib/gitify/app.asar/build/main.js:2:289919
    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:87415)
    at WebContents.emit (node:events:524:28)
11:43:01.615 › [loginWithGitHubApp] failed to login with GitHub Error: Error invoking remote method 'gitify:safe-storage-encrypt': Error: Error while encrypting the text provided to safeStorage.encryptString. Encryption is not available.
    at IpcRenderer.invoke (node:electron/js2c/renderer_init:2:7076)
[...]

PR #2785, opened 2026-04-16
fix: Improve password store detection on Linux tiling WMs

BlueManCZ:
## Summary

On Linux tiling WMs (Hyprland, Sway, i3, etc.), Electron's `safeStorage` API fails to detect the correct password storage backend, even when a keyring like `gnome-keyring` is installed and running. This causes login to fail with `Encryption is not available`.

The root cause is that Electron selects the backend based on the `XDG_CURRENT_DESKTOP` environment variable. When it doesn't match a known desktop environment, it falls back to `basic_text` and reports `isEncryptionAvailable: false` -- despite `org.freedesktop.secrets` being available on D-Bus.

This PR adds a startup check that queries D-Bus for an active Secret Service provider. If one is found and the user hasn't explicitly passed `--password-store`, it sets `--password-store=gnome-libsecret` so Electron uses the correct backend.

## Details

- Only runs on Linux, before the app ready event (where command line switches must be set)
- Uses `execFileSync` to call `dbus-send` -- no shell spawned, no new dependencies
- Wrapped in try/catch with a 2s timeout -- if D-Bus is unavailable, it's a no-op
- Respects any explicit `--password-store=...` flag the user passes

## Related

- #1981 -- "Encryption is not available" on Arch Linux
- #2309 -- "Fail to authenticate in hyprland/wayland"

## Tested on

- Gentoo Linux x86_64 (kernel 6.18.12)
- Hyprland
- gnome-keyring-daemon

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1,6 +1,36 @@
+import { execFileSync } from 'node:child_process';
+
 import { app } from 'electron';
 import log from 'electron-log';
 import { menubar } from 'menubar';
+
+// On Linux tiling WMs (Hyprland, Sway, i3, etc.), Electron fails to detect the
+// correct password storage backend because XDG_CURRENT_DESKTOP doesn't match a
+// known DE. Query D-Bus for an active Secret Service provider and, if found, tell
+// Electron to use gnome-libsecret so safeStorage works correctly.
+if (
+  process.platform === 'linux' &&
+  !app.commandLine.hasSwitch('password-store')
+) {
+  try {
+    execFileSync(
+      'dbus-send',
+      [
+        '--session',
+        '--dest=org.freedesktop.DBus',
+        '--type=method_call',
+        '--print-reply',
+        '/org/freedesktop/DBus',
+        'org.freedesktop.DBus.GetNameOwner',
+        'string:org.freedesktop.secrets',
+      ],
+      { timeout: 2000, stdio: 'ignore' },
+    );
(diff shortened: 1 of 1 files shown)


setchy:
Should this fix be upstream in electron itself?

BlueManCZ:
After digging deeper, I found that Electron has actually addressed this upstream in electron/electron#49054, which closes both electron/electron#39789 and electron/electron#47436.

The fix introduces new async `safeStorage` methods (`encryptStringAsync` / `decryptStringAsync`) that use `os_crypt_async`, which properly discovers secret service providers without relying on `XDG_CURRENT_DESKTOP`. However, it's labeled `no-backport` and `semver/major`, so it only lands in **Electron 42**, which is currently in beta and scheduled for stable on **May 5, 2026**.

The migration on Gitify's side should be minimal — literally changing two lines in `src/main/handlers/storage.ts`, swapping `encryptString` / `decryptString` for their async counterparts. The IPC handlers already support async, so no other changes are needed.

So there are two options:

1. **Merge this D-Bus detection now** and remove it after migration to async `safeStorage` from Electron 42
2. **Wait ~3 weeks** for Electron 42 stable and migrate to the async `safeStorage` API directly

setchy:
Thanks for triaging further. Let's wait for v42

setchy (changes requested):
Appreciate the triage - as commented, let's wait for`electron@42.x.x` to be released

PR #2787, opened 2026-04-18
Add Gitea Notifications

bircni:
Add the option to also add gitea notifications
grafik


setchy (changes requested):
At this time, supporting other git platforms is not on our roadmap (see CONTRIBUTING.md)

afonsojramos:
@setchy to be honest, looking at the implementation (just reviewing the code, haven't really test) it doesn't really seem to create any Gitea-specific concepts. In the past, bitbucket has definitely steered far from what we had for GH, but gitea doesn't seem to differ as much

setchy:
> @setchy to be honest, looking at the implementation (just reviewing the code, haven't really test) it doesn't really seem to create any Gitea-specific concepts. 

Same, I too have done a quick code review but haven't tested it.  Took me a minute to work out what `forge` was being used for 😄 


> In the past, bitbucket has definitely steered far from what we had for GH, but gitea doesn't seem to differ as much

Actually, **Bitbucket-only** notifications _could_ be a clean fit too...  it was just that there are also many other Atlassian products [non-git] which had valuable notifications.

---

I think my primary hesitation is the overhead on the maintainers to support more git platforms.  

In concept, I like it.  In practice, there will be a "cost"... 

It would also make us question whether the use of things like Octicons, Octokit, Primer Design System still makes sense...

afonsojramos:
> It would also make us question whether the use of things like Octicons, Octokit, Primer Design System still makes sense...

Fair, but we are lacking the design skills to do something from scratch 😅 It should be fine

> I think my primary hesitation is the overhead on the maintainers to support more git platforms.

It should be fine if we have someone responsible for that section. In this case, @bircni 👀 Other apps that consider themselves as mutli-platform also have dedicated maintainers for each OS, to validate that everything is working.

> Actually, Bitbucket-only notifications could be a clean fit too... it was just that there are also many other Atlassian products [non-git] which had valuable notifications.

Well then 👀

bircni:
rebased onto main - idk how we want to proceed here 
I really like the app and gitea notifications are a gamechanger for me here

afonsojramos:
@bircni let's just take a step back and evaluate multi-platform, maybe this can be architectured in a different way.

setchy:
https://github.com/gitify-app/gitify/issues/2841

setchy:
thanks @bircni 

now superseded by forge adapters in #2842

PR #2839, opened 2026-05-06
chore(deps): update electron to v42 and migrate to async safeStorage

afonsojramos:
## Summary

- Bumps `electron` from 41.5.0 to 42.0.0.
- Migrates `src/main/handlers/storage.ts` to the new async `safeStorage.encryptStringAsync` / `decryptStringAsync` introduced in [electron/electron#49054](https://github.com/electron/electron/pull/49054), which uses `os_crypt_async` to discover Linux secret service providers via D-Bus instead of relying on `XDG_CURRENT_DESKTOP`.
- Fixes "Encryption is not available" on Linux tiling WMs (Hyprland, Sway, i3, etc.) — supersedes the D-Bus workaround proposed in #2785, which can now be closed.

## Test plan

- [x] \`pnpm install\` clean on Electron 42.0.0
- [x] \`pnpm test\` — 933 tests / 133 files passing
- [x] \`pnpm lint:check\` clean on changed files
- [ ] Manual verification of login flow on Linux Hyprland / macOS / Windows

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "concurrently": "9.2.1",
     "date-fns": "4.1.0",
     "dotenv": "17.4.2",
-    "electron": "41.5.0",
+    "electron": "42.0.0",
     "electron-builder": "26.8.1",
     "final-form": "5.0.0",
     "graphql": "16.13.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -19,7 +19,7 @@ importers:
         version: 6.8.3
       menubar:
         specifier: 9.5.2
-        version: 9.5.2(electron@41.5.0)
+        version: 9.5.2(electron@42.0.0)
       react:
         specifier: 19.2.5
         version: 19.2.5
@@ -145,8 +145,8 @@ importers:
         specifier: 17.4.2
         version: 17.4.2
       electron:
-        specifier: 41.5.0
-        version: 41.5.0
+        specifier: 42.0.0
+        version: 42.0.0
       electron-builder:
         specifier: 26.8.1
         version: 26.8.1(electron-builder-squirrel-windows@26.8.1)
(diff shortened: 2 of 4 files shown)


afonsojramos:
@bluemanCZ can you test this branch to validate that it indeed works?

setchy:
Awesome!!! From memory there may be a website FAQ re Linux flags we can cleanup now, too!

setchy (approved):

BlueManCZ:
@afonsojramos Pulled this down locally and the dev build fails to authenticate — Octokit throws `Token passed to createTokenAuth is not a string`.

The cause is in `src/main/handlers/storage.ts`: `safeStorage.decryptStringAsync` doesn't return a string like the sync `decryptString` did — it returns `{ shouldReEncrypt: boolean, result: string }` ([Electron docs](https://www.electronjs.org/docs/latest/api/safe-storage#safestoragedecryptstringasyncencrypted)). The current handler returns the whole object, so the renderer ends up storing/passing that object as the token.

Minimum fix is to unwrap `.result` before returning. Worth also handling `shouldReEncrypt` — when it's `true`, re-encrypt the token and persist it so we transparently roll forward when the OS keychain rotates keys.

setchy:
> Awesome!!! From memory there may be a website FAQ re Linux flags we can cleanup now, too!

This one: https://gitify.io/faq/#encryption-not-available-linux

afonsojramos:
@BlueManCZ great catch, thanks for pulling it down and digging in. 

You were right, fixed in e1049f51. It now unwraps `.result` now. also tightened the mock typing in storage.test.ts (satisfies `Pick<SafeStorage, ...>`) so this exact drift fails at tsc time next round.

On the re-encrypt-and-persist piece - leaving it as a follow-up since it touches the renderer’s auth store.

BlueManCZ:
Thank you for the changes. I can confirm that it works now without any issues on Hyprland. No need for the `--password-store` flag anymore.

PR #2845, opened 2026-05-06
feat(settings): add "Close to tray" option

BlueManCZ:
## Summary

Adds an opt-in **"Close to tray"** checkbox to **System** settings (default **off**). When enabled, pressing your window manager's close shortcut (e.g. `Super+Q` on Hyprland, `Alt+F4` on Windows, `Cmd+W` on macOS) **hides** the Gitify popup instead of **quitting** the app — the same way Slack, Signal, and most messaging apps behave.

## Why

On Linux tiling WMs (Hyprland, i3, Sway, …) users typically bind a global "close window" shortcut. Today, hitting that shortcut on Gitify's popup terminates the entire app — you lose the tray icon and have to re-launch Gitify to get notifications back. Pressing `Esc` or clicking outside already hides the popup gracefully; this PR makes the WM close shortcut do the same when the user opts in.

The setting is **off by default**, so existing behavior is preserved. Real ways to quit Gitify (tray right-click → "Quit Gitify", `Ctrl/Cmd+Q`) are unchanged and always work regardless of the setting.

## Where it lives

`Settings → System → Close to tray` (right after "Open at startup"), with a tooltip explaining the behavior and how to actually quit.

## Behavior matrix

| Setting | $MOD+Q / $ALT+F4 | `Esc` | Tray "Quit Gitify" | `Ctrl/Cmd+Q` |
|---------|------------------|-------|--------------------|--------------|
| **Off** (default) | quits app | hides popup | quits app | quits app |
| **On**            | hides popup | hides popup | quits app | quits app |

## Implementation notes for reviewers

The two non-obvious bits are documented inline in `src/main/lifecycle/window.ts`. Calling them out here so the diff doesn't look surprising:

1. **Capturing the `BrowserWindow` reference at config time.** `menubar` registers its own `close` listener (`windowClear`) when the window is created. It nulls `mb.window` regardless of `preventDefault`, and listeners run in registration order — so reading `mb.window` inside our handler returns `undefined`. We capture the reference at the top of `configureWindowEvents` and, after hiding, restore menubar's internal field so the next tray click reuses the same window (keeping renderer state — keyboard listeners, notification cache, scroll, etc.). The private-field write is wrapped in a `try/catch` so a future menubar refactor degrades to "fresh window per show" rather than throwing.

2. **Deferring `hide()` via `setImmediate`.** On Wayland, calling `hide()` synchronously after `preventDefault` on a frameless surface can leave the window mapped but without keyboard input routing. Deferring lets the close cancellation unwind first.

3. **`window-all-closed` safety net.** Suppresses the default Electron quit so the tray icon stays put if the WM tears the window down despite `preventDefault` — a known Wayland edge case (electron/electron#34788, #35657). `menubar` then recreates the window on the next tray click.

## Test plan

- [ ] With the setting **off** (default) on Linux: $MOD+Q quits the app — same as `main`.
- [ ] With the setting **on** on Linux: $MOD+Q hides the popup; tray icon stays; clicking the tray reopens the same window with state intact (no flicker, no notification re-fetch).
- [ ] Tray right-click → "Quit Gitify" fully exits regardless of the setting.
- [ ] `Ctrl/Cmd+Q` accelerator fully exits regardless of the setting.
- [ ] `Esc` still hides the popup regardless of the setting.
- [ ] Toggling the setting persists across app restarts.
- [ ] `pnpm test` passes (11 tests for `lifecycle/window.ts`, Settings snapshot updated, no other test changes needed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


setchy:
I'm wondering whether this needs to be an opt-in user setting at all. 

What's your thoughts @afonsojramos

BlueManCZ:
@setchy My thinking was that if some users actually rely on Gitify terminating when the window is closed, making it opt-in wouldn't break their existing workflow. But it may be such a niche use case that enabling it by default would make more sense for the majority of users. I'll leave that call to you.

src/main/lifecycle/window.ts
```diff
@@ -1,17 +1,129 @@
...
+   * Safety net: if the WM tears down the window despite `preventDefault`
+   * (a known Wayland edge case), suppress the default Electron quit so
+   * the tray icon stays put and `menubar` can recreate the window on
+   * the next tray click.
+   */
+  app.on('window-all-closed', () => {
+    if (keepRunningInTray && !isQuitting) {
+      return;
+    }
+    if (process.platform !== 'darwin') {
```
setchy: Let's use the isMacOS util from `./shared/platform`

afonsojramos:
Yeah, agreed, this is the standard behavior for apps with menubar, and it is one that makes sense!

afonsojramos (approved):

PR #2848, opened 2026-05-07
chore(events): remove dead UPDATE_THEME event and onSystemThemeUpdate API

afonsojramos:
## Summary

`EVENTS.UPDATE_THEME` and the `window.gitify.onSystemThemeUpdate` registration shim are dead code:

- No code in `src/main/` ever calls `sendRendererEvent(mb, EVENTS.UPDATE_THEME, …)` — there is no producer.
- No code in `src/renderer/` ever calls `window.gitify.onSystemThemeUpdate(…)` — there is no consumer.
- Only the registration plumbing and its test exist, plus a `vi.fn()` stub in the renderer test setup.

Theme handling currently happens entirely on the renderer side via `@primer/react`'s `useTheme` and the OS-level `prefers-color-scheme` media query. The IPC channel was never wired up.

## Removed

- `EVENTS.UPDATE_THEME` and its `EventContracts` entry.
- `window.gitify.onSystemThemeUpdate` from `src/preload/index.ts`.
- The corresponding test in `src/preload/index.test.ts`.
- The `onSystemThemeUpdate` mock from the renderer test setup.

## Verified

`pnpm tsc --noEmit` clean, `pnpm test` 937/937, `pnpm lint:check` 0 errors.

If OS theme tracking is wanted in future, the right shape is `nativeTheme.on('updated', …)` on the main side, not this listener-without-a-producer.

### DIFF
--- a/src/preload/index.test.ts
+++ b/src/preload/index.test.ts
@@ -66,7 +66,6 @@ interface TestApi {
   tray: { updateColor: (n?: number) => void };
   openExternalLink: (u: string, f: boolean) => void;
   app: { version: () => Promise<string>; show?: () => void; hide?: () => void };
-  onSystemThemeUpdate: (cb: (t: string) => void) => void;
   raiseNativeNotification: (t: string, b: string, u?: string) => unknown;
 }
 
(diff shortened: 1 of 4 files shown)


setchy (approved):
🧹 🧼

PR #2850, opened 2026-05-07
ci(triage): use pull_request_target so labeler can write on fork PRs

afonsojramos:
## Summary

Switches the Triage workflow trigger from `pull_request` back to `pull_request_target` so `multi-labeler` and `action-semantic-pull-request` can write labels and post a commit status on PRs from forks (e.g. #2845). Under `pull_request`, GitHub forces `GITHUB_TOKEN` to read-only for fork PRs, regardless of the workflow's declared `permissions:`, which is why the labeler check has been failing.

## Why this is OK (and reverting #2763 here is acceptable)

#2763 deliberately moved this file to `pull_request` to clear zizmor's `dangerous-triggers` audit. That audit's load-bearing concern is *future drift* — someone later adds a `run:` step or an unpinned `uses:`, and the privileged context becomes RCE without anyone noticing.

In our case the practical control is review: every change to this file goes through PR review by maintainers who have write access, and the workflow itself is structurally minimal — no `actions/checkout`, no `run:` steps, only SHA-pinned actions that read PR metadata via the GitHub API. None of zizmor's listed vectors (argument injection, env injection, file relinking) have a surface here without one of those building blocks being added, and adding them requires explicit maintainer approval.

The trigger now carries an inline `# zizmor: ignore[dangerous-triggers]` directive with the reasoning above written directly into the workflow, so the exception is self-documenting and zizmor's audit stays clean. Verified locally: `zizmor .github/workflows/triage.yml` → `No findings to report`.

## Test plan

- [x] Zizmor run passes
- [x] On the next fork PR, the `Auto-label PR` check succeeds and labels are applied

### DIFF
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -1,7 +1,7 @@
 name: Triage PR
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
       - main
     types:
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -1,7 +1,16 @@
 name: Triage PR
 
 on:
-  pull_request_target:
+  # `pull_request_target` is required so the labeler/title-validator can
+  # write labels and statuses on PRs from forks (under `pull_request`,
+  # GitHub forces GITHUB_TOKEN to read-only for fork PRs). Safe here
+  # because this workflow:
+  #   - never checks out PR code (no actions/checkout),
+  #   - has no `run:` steps that interpolate PR fields,
+  #   - only invokes SHA-pinned actions that read PR metadata via the API,
+  #   - is locked behind required code-owner review (see .github/CODEOWNERS)
+  #     so future edits cannot quietly add privileged execution surface.
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
     branches:
       - main
     types:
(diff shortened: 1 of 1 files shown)


setchy:
This was our workflow config prior to https://github.com/gitify-app/gitify/pull/2763 where we addressed some feedback from zimzor / actionlint

setchy:
https://docs.zizmor.sh/audits/#dangerous-triggers

> _Many online resources suggest that pull_request_target and other dangerous triggers can be used securely by ensuring that the PR's code is not executed, but this is not true: an attacker can often find ways to execute code in the context of the target repository, even if the workflow doesn't explicitly run any code from the PR. Common vectors for this include argument injection (e.g. with xargs), environment injection (e.g. LD_PRELOAD), and local file inclusion (e.g. relinking files to the runner's credentials file or similar)._

afonsojramos:
Ah, I see 🤔 I'll look at what the best alternative is. Because otherwise external PRs just don't work, which defeats the purpose: https://github.com/gitify-app/gitify/pull/2845

afonsojramos:
Okay, so after a brief investigation, it seems the goal of zizmor’s dangerous-triggers audit here is to warn against future drift — i.e. someone later adding a run: step or unpinned uses: and the privileged context silently becoming an RCE surface — rather than flagging the current file as exploitable. In our case that drift can’t land without a maintainer-approved diff, and the workflow today has no checkout, no run:, and only SHA-pinned actions reading PR metadata via the API. Pushed a follow-up that adds an inline # zizmor: ignore[dangerous-triggers] with that reasoning written directly above the trigger so the exception is self-documenting.

setchy (approved):

PR #2869, opened 2026-05-09
chore: migrate to vite-plus unified toolchain

afonsojramos:
## Summary
- Replace `@biomejs/biome` with [Vite+](https://viteplus.dev/) (`vite-plus@0.1.20`), which bundles `oxlint`, `oxfmt`, Vitest, and the dev/build pipeline behind a single `vp` CLI.
- Consolidate linter, formatter, and pre-commit configuration into the `lint`, `fmt`, and `staged` blocks of `vite.config.ts` — `.oxlintrc.json`, `.oxfmtrc.json`, and the `lint-staged` block in `package.json` are gone.
- Switch package scripts to `vp dev` / `vp build` / `vp test` / `vp check`, swap husky's pre-commit to `vp staged`, and point VS Code's oxc extension at `vite.config.ts`.
- Convert all `biome-ignore` directives to `oxlint-disable`, fix the lint findings oxlint surfaced (constant-binary-expression, unsafe optional chaining, unused param, useNavigate stability), regenerate one snapshot affected by oxfmt's JSX whitespace, and apply oxfmt formatting across the codebase.

## Test plan
- [x] `pnpm check` (vp lint + vp fmt: clean across 339 lint files / 373 fmt files)
- [x] `pnpm tsc --noEmit`
- [x] `pnpm test` (1031/1031 passing)
- [x] `pnpm build`


setchy:
No vite+?

package.json
```diff
@@ -128,13 +128,16 @@
     "vite-plugin-checker": "0.13.0",
     "vite-plugin-electron": "0.29.1",
     "vite-plugin-electron-renderer": "0.14.7",
+    "vite-plus": "0.1.20",
```
setchy, line 131: Does vite-plus mean we should drop vite and vitest as direct devdeps?

vite.config.ts
```diff
@@ -98,95 +89,150 @@ const reactDevToolsPlugin = (): Plugin => ({
...
+  // During CI builds `process.env.OAUTH_CLIENT_ID` will be injected via the environment.
+  define: isBuild
+    ? {
+        'process.env.OAUTH_CLIENT_ID': JSON.stringify(process.env.OAUTH_CLIENT_ID ?? ''),
+      }
+    : {
+        // Development Keys - See CONTRIBUTING.md
+        'process.env.OAUTH_CLIENT_ID': JSON.stringify('Ov23liQIkFs5ehQLNzHF'),
+      },
+  lint: {
```
setchy, line 174: It'd be good to port the biome.json import grouping+sorting rules across, too

https://oxc.rs/docs/guide/usage/formatter/sorting

afonsojramos: Done in 05bf083c — ported biome's groupings (Node builtin → react/testing → electron stack → primer → octokit → other externals → mocks/helpers → shared → constants → state/routing → ui → types → relative) to oxfmt's sortImports with customGroups + newlinesBetween. The only re-sorted call sites were App.tsx and vitest.setup.ts (794c57e1).

package.json
afonsojramos: In principle yes — vite-plus aliases vite (@voidzero-dev/vite-plus-core) and vitest (@voidzero-dev/vite-plus-test) and the docs recommend dropping the standalone deps after rewriting imports to vite-plus / vite-plus/test.

I tried it and the rewrite forces tsconfig moduleResolution to switch from "Node" to "Bundler" so TS can resolve the subpath exports. Bundler resolution then exposes pre-existing latent issues that Node resolution masked:
- src/renderer/types.ts imports from @primer/react/dist/Button/types (deep import not in @primer/react's exports map)
- src/renderer/utils/ui/theme.ts imports from @primer/react/dist/ThemeProvider (same)
- @testing-library/jest-dom matchers (toBeInTheDocument, toHaveTextContent) stop being picked up via vitest/globals types and need to be wired through vite-plus/test/globals or via setupFiles types

Each is fixable but the diff balloons well beyond the toolchain swap. I reverted those changes and would rather chase the cleanup in a follow-up.

afonsojramos: Took another pass. Two of the three issues are fixable cheaply, but dropping vite or vitest themselves still hits ecosystem walls:

~Fixed in cfcfa5ac (regardless of devdep cleanup):~
~- src/renderer/types.ts and utils/ui/theme.ts now inline VariantType and ColorModeWithAuto so we no longer reach into @primer/react/dist/*~

> Ended up reverting the above. Does not feel like it is worth it at the moment.

Conclusion, could not drop vite from devdeps: vite-plugin-electron does require('vite') at the project root. With vite removed, pnpm only places vite-plus-core's vite alias inside vite-plus-core itself, so the plugin's resolve fails. Re-adding vite restores the build.

Could not drop vitest from devdeps either: vitest's matchers are augmented via @testing-library/jest-dom/vitest, which does declare module 'vitest'. Once vitest is gone the global expect is typed as @voidzero-dev/vite-plus-test's Assertion (declared inside dist/chunks via a relative module spec that isn't reachable from outside the package's exports map). I tried augmenting via the package, the dist subpath, and tsconfig paths; none of them merge with the inner declaration. Keeping vitest as a devdep keeps the existing augmentation working.

setchy (approved):

PR #2875, opened 2026-05-10
feat(settings): add "Keep window open when it loses focus" option

BlueManCZ:
## Summary

Adds an opt-in **"Keep window open when it loses focus"** checkbox to **System** settings (default **off**). When enabled, the Gitify popup stays visible after focus moves to another window — useful when triaging notifications side-by-side with the browser, an editor, or a chat client.

## Why

By default, Electron menubar apps hide the popup as soon as it loses focus (the `menubar` library's blur-driven hide). That matches the glance-and-go menubar convention but actively gets in the way of side-by-side workflows: read a PR in the browser while keeping the notification list visible, drag-and-drop URLs into another tool, etc. On tiling WMs (Hyprland, i3, Sway) where there's no fixed "menubar" zone, the convention itself feels out of place.

The setting is **off by default**, preserving existing behavior.

## Where it lives

At the bottom of `Settings → System`, after the existing checkboxes. One-line tooltip explains the behavior.

## Implementation notes for reviewers

1. **The mechanism is purely JS-level.** `menubar`'s blur listener short-circuits its own hide when the underlying `BrowserWindow.isAlwaysOnTop()` returns true (`node_modules/menubar/lib/Menubar.js:343-353`). We toggle that flag from the renderer-driven setting via a new IPC event (`UPDATE_KEEP_WINDOW_ON_BLUR`). No window-manager hints or platform-specific paths involved — works uniformly on macOS, Windows, X11, and (verified) Hyprland.

2. **Small bug fix folded in.** The existing `devtools-opened` handler in `src/main/lifecycle/window.ts` sets `alwaysOnTop=true` for DevTools usability. The matching `devtools-closed` handler previously did **not** restore it, so opening and closing DevTools left the flag stuck on for the rest of the session. The new `applyKeepWindowOnBlur` setter owns this flag's lifecycle, and the `devtools-closed` handler now restores it to the user's preference (`false` by default, matching the prior intent). One line of behavior change, inseparable from owning the flag — calling it out so the diff doesn't look surprising.

3. **Naming across the IPC boundary.** Renderer concept: `keepWindowOnBlur`. Main-side setter: `applyKeepWindowOnBlur(mb, value)`. The fact that we implement it via `setAlwaysOnTop` is an implementation detail and not exposed in the public name.

## Behavior matrix

| Setting | Click outside window | `Esc` | Tray icon click | Open DevTools → close DevTools |
|---------|---------------------|-------|-----------------|-------------------------------|
| **Off** (default) | hides popup | hides popup | toggles | flag returns to `false` ✓ (was previously stuck on) |
| **On** | popup stays | hides popup | toggles | flag returns to `true` ✓ |

## Test plan

- [ ] With setting **off** (default): clicking outside the window hides it — same as `main`.
- [ ] With setting **on**: clicking outside leaves the popup visible; `Esc` still hides it; tray icon click still toggles.
- [ ] Setting persists across app restart.
- [ ] Open DevTools, then close it — `alwaysOnTop` returns to the setting's current value (regression check on the bug fix above).
- [ ] `pnpm test` — 1038 tests pass; new tests cover `applyKeepWindowOnBlur`, `devtools-closed` restoration, the settings checkbox, and the IPC handler; Settings snapshot regenerated.
- [ ] `pnpm check` and `pnpm tsc --noEmit` clean.
- [ ] Manually verified on Hyprland (Wayland). Mechanism is JS-level so other platforms should behave identically — I don't have macOS or Windows to verify locally.

### DIFF
--- a/src/renderer/components/settings/SystemSettings.tsx
+++ b/src/renderer/components/settings/SystemSettings.tsx
@@ -297,15 +297,6 @@ export const SystemSettings: FC = () => {
         </Stack>
 
         <Checkbox
-          checked={settings.openAtStartup}
-          label="Open at startup"
-          name="openAtStartup"
-          onChange={() => updateSetting('openAtStartup', !settings.openAtStartup)}
-          tooltip={<Text>Launch {APPLICATION.NAME} automatically at startup.</Text>}
-          visible={!window.gitify.platform.isLinux()}
-        />
-
-        <Checkbox
           checked={settings.keepWindowOnBlur}
           label="Keep window open when it loses focus"
           name="keepWindowOnBlur"
@@ -316,6 +307,15 @@ export const SystemSettings: FC = () => {
               it.
             </Text>
           }
+        />
+
+        <Checkbox
+          checked={settings.openAtStartup}
+          label="Open at startup"
+          name="openAtStartup"
+          onChange={() => updateSetting('openAtStartup', !settings.openAtStartup)}
+          tooltip={<Text>Launch {APPLICATION.NAME} automatically at startup.</Text>}
+          visible={!window.gitify.platform.isLinux()}
         />
       </Stack>
     </fieldset>
--- a/src/renderer/components/settings/SystemSettings.tsx
+++ b/src/renderer/components/settings/SystemSettings.tsx
@@ -304,6 +304,19 @@ export const SystemSettings: FC = () => {
           tooltip={<Text>Launch {APPLICATION.NAME} automatically at startup.</Text>}
           visible={!window.gitify.platform.isLinux()}
         />
+
+        <Checkbox
+          checked={settings.keepWindowOnBlur}
+          label="Keep window open when it loses focus"
+          name="keepWindowOnBlur"
+          onChange={() => updateSetting('keepWindowOnBlur', !settings.keepWindowOnBlur)}
+          tooltip={
+            <Text>
+              Prevent the {APPLICATION.NAME} window from automatically hiding when you click outside
+              it.
+            </Text>
+          }
+        />
       </Stack>
     </fieldset>
   );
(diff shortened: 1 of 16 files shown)


setchy:
Isn't this already covered by our Open Links system preference...

BlueManCZ:
No, it is not. Gitify still closes when the window loses focus, no matter what I set there. This has nothing to do with the links. Does this behave differently on other platforms?

BlueManCZ:
Just to show what I mean. Note that I'm using only mouse and the focus follows the cursor.

https://github.com/user-attachments/assets/ab86a89d-6bce-44ed-a761-3d4aedb6210a

src/renderer/components/settings/SystemSettings.tsx
```diff
@@ -304,6 +304,19 @@ export const SystemSettings: FC = () => {
           tooltip={<Text>Launch {APPLICATION.NAME} automatically at startup.</Text>}
           visible={!window.gitify.platform.isLinux()}
         />
+
+        <Checkbox
```
setchy, line 299: Can we please move this above the openAtStartup option

setchy:
> Just to show what I mean. Note that I'm using only mouse and the focus follows the cursor.
> 
> https://github.com/user-attachments/assets/ab86a89d-6bce-44ed-a761-3d4aedb6210a
> 

Ty

Aside: the window positioning really is proper broken on Linux🤢

afonsojramos:
> the window positioning really is proper broken on Linux🤢

Is that not a hyprland issue specifically?

src/renderer/components/settings/SystemSettings.tsx
BlueManCZ: Done

BlueManCZ:
> > the window positioning really is proper broken on Linux🤢
> 
> Is that not a hyprland issue specifically?

Yes, this is Hyprland specific. Tried it on Gnome and there it opens correctly near the tray icon.

PR #2880, opened 2026-05-11
refactor(main): adopt `electron-menubar` v10 for window/tray/system handling

afonsojramos:
## Summary
- Swap the upstream `menubar` dependency for our hard forked `electron-menubar` (v10) and lean on its built-in features.
- Delegate window lifecycle (`hideOnClose`, `escapeToHide`, `recenterOnTray`), tray context-menu wiring/SNI refresh, and the global keyboard shortcut to the menubar instance.
- Drop the now-redundant local glue in `lifecycle/window.ts`, `handlers/tray.ts`, and `handlers/system.ts`, simplifying tests accordingly.


setchy:
These comments are more for `@gitify/menubar`
- should we add `.github/release.yml` to https://github.com/gitify-app/menubar and generate corresponding `v10.0.0` Release Notes?
- should we refresh/update https://github.com/gitify-app/menubar/blob/main/PLATFORMS.md?
- could the three screenshot examples in README be generalized to Linux, Windows and macOS without the old version numbers? 
- update the Electron compatibility table - https://github.com/gitify-app/menubar/blob/main/PLATFORMS.md

setchy:
What do you think about a more universal package name such as `electron-menubar` vs `@gitify/menubar`.  

We could also contribute a Renovate replacement rule which would help all the other consumers of max-mapper/menubar migration from the unmaintained v9 to v10

afonsojramos:
`electron-menubar` is actually already taken :/

setchy:
Oh really, I had searched npmjs before but couldn't see any conflicts

setchy (approved):

PR #2874, opened 2026-05-10
refactor: tidy up the forge adapter interface

afonsojramos:
## Summary

Three follow-up cleanups on top of the forge adapter interface, each as its own commit:

- **Move GitHub-specific auth helpers** (`extractHostVersion`, `getGitHubAuthBaseUrl`, `getDeveloperSettingsURL`, `getNewTokenURL`, `getNewOAuthAppURL`, `isValidToken`, `isValidClientId`) out of shared `auth/utils.ts` and into a new `forges/github/auth.ts`. Tests followed.
- **Drop the `answeredDiscussion` capability** from the shared `ForgeCapabilities` contract. It only ever drove GitHub GraphQL query construction, so it now lives privately as `supportsAnsweredDiscussion` inside `forges/github/capabilities.ts`. `isAnsweredDiscussionFeatureSupported` removed from `api/features.ts` along with the Gitea stub.
- **Rename `getDeveloperSettingsUrl` → `getAccountSettingsUrl`** on the `ForgeAdapter` interface (and `openDeveloperSettings` → `openAccountSettings` on the consumer). The "developer settings" wording was GitHub-flavoured; the new name reads cleanly for any forge.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm exec biome check` on touched files — only pre-existing warnings
- [x] `pnpm exec vitest --run` — 1040/1040 pass

Closes #2873

PR #2882, opened 2026-05-12
refactor(forges): expose participatingDocsUrl on the adapter

afonsojramos:
## Summary

Another follow-up to #2874. The "see official docs" link in NotificationSettings opened a hardcoded GitHub docs URL via a dedicated helper, but the button was shown for every forge — including Gitea-only setups where the link is irrelevant.

- Add `participatingDocsUrl?: Link` to `ForgeAdapter`. Optional because not every forge has equivalent docs.
- Wire `Constants.GITHUB_DOCS.PARTICIPATING_URL` onto `githubAdapter`.
- `NotificationSettings` picks the first signed-in forge's URL and hides the button when none expose one.
- Drop the dedicated `openGitHubParticipatingDocs` helper from `system/links.ts` plus its test.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm lint` — clean
- [x] `pnpm exec vitest --run src/renderer/utils/system/links.test.ts src/renderer/components/settings/NotificationSettings.test.tsx` — passes

### DIFF
--- a/src/renderer/components/settings/NotificationSettings.test.tsx
+++ b/src/renderer/components/settings/NotificationSettings.test.tsx
@@ -212,9 +212,7 @@ describe('renderer/components/settings/NotificationSettings.tsx', () => {
     const tooltipElement = screen.getByLabelText('tooltip-showOnlyParticipating');
 
     await userEvent.click(tooltipElement);
-    await userEvent.click(
-      screen.getByTitle('Open GitHub documentation for participating and watching notifications'),
-    );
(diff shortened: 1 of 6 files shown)


setchy:
Instead of this, I think it might be a good time to explore how Forge-specific Settings are organized...

afonsojramos:
@setchy Fair, but interesting... Not sure how it would work 🤔 Tabbed? Or a section per forge with select supported features?

setchy:
I'm not too certain either... yet...

How about we start with an Issue to elaborate which current settings are forge specific or app generic

PR #2886, opened 2026-05-12
refactor(notifications): generate notification URL via the adapter

afonsojramos:
## Summary

Follow-up to #2874. `notifications/url.ts` imported `getHtmlUrl` directly from `forges/github/client` and called it for every notification — including ones from Gitea accounts. That was a sharp adapter leak.

- Replace the direct `getHtmlUrl` import with `adapter.followUrl<{ html_url: string }>(account, url)`. The Gitea adapter already implements `followUrl` (via `giteaGetJson`), so this works for both forges.
- Rename `generateGitHubWebUrl` → `generateNotificationWebUrl` to match what it actually does.
- Drop the now-unused `getHtmlUrl` helper from `forges/github/client.ts` plus its test, and the `GitHubHtmlUrlResponse` type.
- Tests now spy on `githubAdapter.followUrl` instead of the deleted helper.

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm lint` — clean
- [x] `pnpm exec vitest --run src/renderer/utils/notifications/url.test.ts src/renderer/utils/forges/github/client.test.ts src/renderer/utils/system/links.test.ts src/renderer/utils/system/native.test.ts` — passes

### DIFF
--- a/src/renderer/utils/forges/github/client.test.ts
+++ b/src/renderer/utils/forges/github/client.test.ts
@@ -18,7 +18,6 @@ import {
   fetchPullByNumber,
   getCommit,
   getCommitComment,
-  getHtmlUrl,
   getRelease,
   ignoreNotificationThreadSubscription,
   listNotificationsForAuthenticatedUser,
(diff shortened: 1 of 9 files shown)


setchy (changes requested):
Octokit client has inbuilt caching the typing. I wouldn't want to stray from that

setchy:
> Octokit client has inbuilt caching the typing. I wouldn't want to stray from that

My apologies, reviewed the code closer and I see its still using octokit client

setchy (approved):

PR #2959, opened 2026-06-03
feat(filters): distinguish author and commenter in search filters

### LINKED ISSUE #2914: author: exclude filter suppresses review_requested notifications on matching PRs

### 📝 Description

When using the search filter with `author:somebot` as an exclude term, notifications with reason `review_requested` are hidden if the bot is resolved as the notification's `subject.user` during enrichment, even when the bot did not open the PR.

I have `org:myorg` as an include filter and `author:coderabbitai` as an exclude filter. A coworker requests my review on a human-authored PR. CodeRabbit automatically comments on the PR (as bots do). Since the enrichment handler fetches only the last comment (`LAST_COMMENTS: 1` in `constants.ts`), and `pullRequest.ts` prioritizes the last comment's author over the PR author when setting `subject.user`:

```typescript
const prComment = pr.comments?.nodes?.[0];
const prUser = getNotificationAuthor([prComment?.author, pr.author]);
```

...`subject.user.login` becomes `coderabbitai`. The exclude filter then matches and the notification is filtered out, even though:
1. The PR was opened by a human
2. The notification reason is `review_requested` (someone directly asked for my review)
3. CodeRabbit's only involvement was leaving an automated comment

The `author:` qualifier in `search.ts` extracts `n.subject?.user?.login` unconditionally, and `filter.ts` applies the exclude without considering `notification.reason`. This means you can't filter out bot comment noise without also losing actionable review requests, assignments, and mentions on any PR where the bot happened to leave the last comment.

### 🪜 Steps To Reproduce

1. Enable detailed notifications in settings
2. Add an include search filter: `org:<your-org>`
3. Add an exclude search filter: `author:<some-bot>` (e.g. `author:coderabbitai`)
4. Have a human open a PR in the org where the bot automatically comments
5. Have someone request your review on that PR
6. The "review requested" notification appears on github.com/notifications but not in Gitify

### 🪵 Log Excerpts

_No response_

### Gitify Version

6.17.0

### Operating System

macOS

### GitHub Account

GitHub Cloud

### 📸 Screenshots

_No response_

afonsojramos:
## Summary

The `author:` search filter matched `subject.user`, but enrichment sets `subject.user` to the latest commenter (for the avatar), not the thread author. So `author:coderabbitai` as an exclude hid review requests on human-authored PRs whenever the bot left the last comment, and the filter named "author" was really matching commenters.

This splits that conflated identity into two distinct, forge-agnostic roles on the notification subject:

- `author`: who created the thread (pull request / issue / discussion / release / commit)
- `commenter`: who left the latest comment, when there is one

`subject.user` stays as the derived display identity (`commenter ?? author`), so the avatar is unchanged. Filtering now keys on the roles:

- `author:` matches the thread author
- `commenter:` (new) matches the latest comment author
- the **User Type** filter now keys on the author too, so "Bot" means "authored by a bot" (dependabot, renovate) rather than "a bot left the latest comment"

The filter layer no longer inspects GitHub notification reasons, so it stays generic across forges. For commit notifications the commit author and the latest comment author are fetched in parallel, so both roles are populated without extra latency.

Closes #2914


setchy:
I've been thinking about this FR...

GitHub UI doesn't distinguish between author or commenter, since notifications always show the latest state - so I was tempted to close the FR as unplanned...

That said, alternatively, we could look to change the filter include/exclude tokens to support `author:` and `commenter:` prefixes which could search on the two different user attributes.

I'm not sure I'm a fan of changing the filter based on the notification subject code

afonsojramos:
@setchy fair... It is also forge specific code that is leaking A LOT tbh

afonsojramos:
Reworked this to drop the reason-based approach and go with the `author:` / `commenter:` split you suggested, @setchy.

Enrichment now exposes two distinct roles on the subject (`author` for the thread creator, `commenter` for the latest comment author), and the filter qualifiers read those directly instead of branching on the notification reason. `subject.user` stays as the display identity (`commenter ?? author`), so avatars and the user-type filter are unchanged. The filter layer no longer touches GitHub reasons, so the forge-specific leak is gone too.

Net effect:
- `author:` matches the actual author now, so excluding a bot no longer suppresses review requests where it only commented last (the original report).
- `commenter:` is the new lever for the bot comment noise case.

Two caveats with the current implementation, however: 
1. `commenter:` is a hammer on a screw, since we no longer reason about why a notification arrived it will also hide a human PR where the bot happened to comment last even if you were review requested. 
2. Commit notifications only populate one of the roles, since we fetch either the comment or the commit, not both.

afonsojramos:
Followup: managed to fix the commit caveat. Commit notifications now fetch the commit and its latest comment in parallel, so both author and commenter are populated like the other types

afonsojramos:
So, final caveat: If one excludes commenter:coderabbitai, a teammate opens a PR and requests a review, then CodeRabbit auto-comments after the request and becomes the latest commenter. That review-request notification gets hidden even though you were asked to review. But maybe this is a disadvantage of excluding bot comments? Or an incentive to be selective of which commenters to exclude. But, to be honest, this was already the case, so nothing new.

**Before:** author:<bot> was the only tool, it secretly matched the last commenter, and you had no way to avoid that. Every user of that filter hit the bluntness.

**Now:** author:<bot> matches the real author, so it’s precise and no longer drops review requests. The old blunt “match whoever commented last” behavior still exists, but only if you deliberately reach for commenter:<bot>.

setchy:
Thanks @afonsojramos.

Let me soak on this a bit. 

We also have a Bot filter which includes dependabot, renovate, netlify. I wonder if we should expand that

afonsojramos:
@setchy Went ahead and applied the same split to the User Type filter too, check it out. “Bot” now keys on the thread author, so it means “authored by a bot” (dependabot, renovate) rather than “a bot left the last comment”. Same fix as the author: qualifier, just on the type toggle.

Only flag is it does change how that filter behaves, so we can always add a new separate toggle for this in particulr

afonsojramos:
Merging this. We can undo the change if the community does not like this filtering workflow

PR #3029, opened 2026-07-08
feat(notifications): display who's the author

yuriteixeira:
## Summary
- Include issue and pull request authors in the notification reason title text
- Keep author names out of the visible footer layout to avoid crowding row actions
- Update notification footer tests and snapshots

## Validation
- pnpm exec vp test --changed --passWithNoTests --update
- pnpm check


setchy:
I don't think we have enough horizontal screen real estate to show notification author handles, as that'll truncate our notification metric pills.

setchy:
Do you have some screenshots of what this feature will look like?

yuriteixeira:
Fair point, I will look into a way to better display this and will get back to you.

PR #3031, opened 2026-07-10
refactor(auth): remove legacy plaintext token migration

afonsojramos:
## Summary

- Removes the temporary plaintext → encrypted token migration that ran on startup (marked with a TODO for a future release).
- Keeps keychain key-rotation persistence under `persistRotatedAuthTokens`.
- Renames `migrateLegacyAuthState` to `sanitiseAuthState` so permanent auth sanitisation is not framed as a one-shot migration.
- Updates startup tests: decrypt failures no longer trigger re-encryption.

**Breaking change:** users who still have unencrypted tokens from pre-safeStorage releases will need to re-login. Anyone who already upgraded through an encrypting release is unaffected.

## Test plan

- [x] `pnpm exec vitest run src/renderer/context/App.test.tsx` (19 tests pass)
- [ ] Manual: app with already-encrypted tokens still logs in and refreshes accounts on startup
- [ ] Manual: keychain rotation still persists rotated ciphertext when `reEncryptedToken` is returned


setchy:
Great, thanks for picking this up mate

setchy (approved):

PR #3066, opened 2026-07-23
fix: stop notification polling from multiplying API requests

### LINKED ISSUE #3065: Rate limit hit with v7

### 📝 Description

I upgraded from v6.20.0 to v7.0.0 yesterday and 2 hours later I got called by my infra team because I was making 15k+ more connection to my GitHub Enterprise server.

I did not change anything in the config, so I could _suspect_ the v7 might have some weird behaviour (which I will try to reproduce)

### 🪜 Steps To Reproduce

TBD 

### 🪵 Log Excerpts

_No response_

### Gitify Version

7.0.0

### Operating System

macOS

### GitHub Account

Combination

### 📸 Screenshots

_No response_

afonsojramos:
## Summary

Upgrading to v7 could multiply notification API traffic by an order of magnitude against GitHub/GHES (15k+ extra connections reported). The TanStack Query refactor made every `useNotifications` consumer (each notification row, repo group, sidebar, header, filter panel) a query observer, and TanStack Query schedules `refetchInterval` **per mounted observer, not per query**. Each observer polled on its own staggered timer, so their fetches never deduped, and `refetchIntervalInBackground: true` kept all of them firing while the menubar sat hidden. A full notification list produced dozens of complete fetch + GraphQL-enrichment cycles per interval instead of one.

Three fixes, in commit order:

- **Singleton polling ownership** — `refetchInterval`, `refetchOnMount`, `refetchOnReconnect` and `refetchOnWindowFocus` are now gated behind the existing `withSideEffects` flag, so only the app-level `GlobalEffects` host schedules refetches. All other consumers share the cached data, exactly as they already did for sound/native alerts.
- **IPC listener leaks** — the same per-consumer pattern registered a permanent `SYSTEM_WAKE` listener per mounted row (the preload bridge offered no unsubscribe at all), piling up stale closures that each fired a refetch on every wake; the OAuth web flow likewise leaked one `AUTH_CALLBACK` listener per login attempt. Preload event registrations now return unsubscribe functions, effects clean up on unmount, the wake listener is host-only, and the auth listener is removed once the flow settles.
- **Retry cooldown** — a failed poll retried near-instantly (v6 never retried within a poll), doubling request volume against an already-struggling server. The single retry now waits 30 seconds.

## Validation

- New regression test mounts a side-effects host plus staggered plain consumers and asserts one fetch per interval; against the old code it fails with 7 fetches where 4 are expected, reproducing the multiplication.
- New tests assert plain consumers register zero wake listeners and that preload unsubscribe actually removes the `ipcRenderer` listener.
- All touched-file suites pass (52 tests); `vp check` (format + lint + tsc) clean.

Closes #3065


src/renderer/hooks/useNotifications.ts
```diff
@@ -154,9 +160,12 @@ export const useNotifications = ({
 
     placeholderData: keepPreviousData,
 
-    refetchInterval: fetchType === FetchType.INTERVAL ? fetchIntervalMs : false,
-    refetchOnReconnect: true,
-    refetchOnWindowFocus: true,
+    // Only the singleton side-effects host polls. Other consumers share the
+    // cached data and would otherwise each schedule their own refetch timer.
+    refetchInterval: withSideEffects && fetchType === FetchType.INTERVAL ? fetchIntervalMs : false,
```
setchy, line 165: A comment for future us... I was thinking we can drop the FetchType setting now that we're using Tanstack Query since it's better at managing the fetch interval, background fetch, fetch on reconnect, etc

setchy:
Interesting - I wonder why/how I didn't have this within Atlassify.  Must be handling this in a different way.

PR #3082, opened 2026-07-26
fix: prevent Windows tray blur→hide race on window show (#3064)

### LINKED ISSUE #3064: Unable to open window

### 📝 Description

I just updated to latest v7.0.0 version published yesterday (previously on latest 6.x version), and I can no longer open the app window when I click on the tray icon. Nothing happens as well when I right-click on the icon. I have to use the Task Manager to close the app.

It seems the extension still work though, as I received notifications when the app started and the icon is properly colored.

### 🪜 Steps To Reproduce

Update to 7.0.0 version

### 🪵 Log Excerpts

_No response_

### Gitify Version

7.0.0

### Operating System

Windows

### GitHub Account

GitHub Cloud

### 📸 Screenshots

_No response_

marceli1404:
## Problem

Issue #3064: On Windows v7.0.0/v7.0.1, clicking the tray icon produces no visible result. The app runs in background (notifications, colored icon) but the window never appears. Silent failure — no errors in logs.

## Root Cause

\lectron-menubar\'s internal blur handler hides the window 100ms after it loses focus. On Windows, clicking the tray icon may not transfer focus to the window, causing the blur→hide timeout to fire immediately after \showWindow()\. When this timeout fires before the first paint cycle, the user sees no visible result.

Relevant Menubar.ts code:
\\\	ypescript
this._browserWindow.on('blur', () => {
  if (!this._browserWindow) return;
  this._browserWindow.isAlwaysOnTop()
    ? this.emit('focus-lost')
    : (this._blurTimeout = setTimeout(() => {
        this.hideWindow();
      }, 100));
});
\\\

If \lwaysOnTop\ is true, the blur handler emits \ocus-lost\ instead of hiding — this is the escape hatch.

## Fix

Listen for the \fter-show\ event and temporarily set \lwaysOnTop=true\ for 400ms after showing the window. This causes the blur handler to short-circuit via the \ocus-lost\ path instead of hiding. The user's \keepWindowOnBlur\ preference is restored after the blur race window passes.

Fixes #3064

### DIFF
--- a/src/main/lifecycle/window.ts
+++ b/src/main/lifecycle/window.ts
@@ -1,7 +1,7 @@
 import { app } from 'electron';
 import type { Menubar } from 'electron-menubar';
 
-import { isMacOS } from '../../shared/platform';
+import { isMacOS, isWindows } from '../../shared/platform';
 
 import { WindowConfig } from '../config';
 import type MenuBuilder from '../menu';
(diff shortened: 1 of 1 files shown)


afonsojramos (approved):

afonsojramos:
I was not quite able to reproduce on the windows machines I had access to, but hopefully this got addressed in the `electron-menubar` dependency.

marceli1404:
Thanks for checking, @afonsojramos. The fix hooks into the fter-show event and temporarily sets lwaysOnTop: true for 400ms so the blur handler emits ocus-lost instead of calling hideWindow(). This works regardless of whether lectron-menubar upstream has been updated, since it patches the behavior at the BrowserWindow level.

If you have a Windows machine available to test, the key scenario is: left-click or right-click the tray icon — the window should appear and remain visible instead of silently disappearing. Without the fix, the window flashes briefly and vanishes.

afonsojramos:
We are the maintainers of the dependency, so the fix is easy to apply :)

setchy:
Thanks everyone! 🙇 A good workaround. 

I'm hopeful that electron fixes this themselves in a future release. I recall reading something about windows blur behavior changing, but wasn't expecting this type of regression.

PR #3100, opened 2026-07-28
fix: re-hide macos dock icon after parentless dialogs

### LINKED ISSUE #3069: macOS:  `dock` icon occasionally showing
I've noticed that the Gitify macOS dock icon occasionally is showing since v7.0.0 update.

Restarting the app restores it to it's `always-hidden` behavior

afonsojramos:
## Summary

Since v7.0.0 the macOS dock icon occasionally reappears even though Gitify runs as a dock-hidden (accessory) app, and only a restart re-hides it.

## Investigation

Tested live on macOS by driving the real built app (hidden window, `setVisibleOnAllWorkspaces`, accessory policy) and measuring `app.dock.isVisible()`:

- **Reproduced:** `app.dock.hide()` can be **silently dropped** when it races an app activation transition. In the experiment, `hide()` called 500ms after the dock became visible left the icon showing; a later `hide()` succeeded. `electron-menubar` calls `dock.hide()` very early in startup, so a lost race there leaves the icon stuck for the whole session. A restart re-rolls the race, which matches "occasionally showing" + "restart restores it".
- **Not reproduced (on this machine/OS):** the parentless dialogs (updater "restart to install", first-run prompt) did *not* flip the dock. Kept the dialog-path re-hides anyway as cheap defense, since dialog-driven activation is version-dependent macOS behavior.
- DevTools (detached) also did not affect the dock.

## Fix

1. **Startup re-assert** — moved into `electron-menubar` per review feedback, since the library owns the `showDockIcon: false` contract: gitify-app/electron-menubar#123, released as `10.1.4`. This PR bumps the dependency and drops the app-level copy.
2. **Dialog-close re-hides** (app-level, defensive): re-assert hidden after the updater dialog is dismissed and after the first-run prompt closes.

## Tests

- `updater.test.ts`: dismissing the update dialog re-hides the dock.
- `first-run.test.ts`: the dock is re-hidden after the prompt closes.
- Library-side coverage for the startup race lives in electron-menubar#123.

Full suite passes (1198 tests), `pnpm check` and `tsc --noEmit` clean.

Closes #3069

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
   },
   "dependencies": {
     "electron-log": "5.4.4",
-    "electron-menubar": "10.1.3",
+    "electron-menubar": "10.1.4",
     "electron-updater": "6.8.9",
     "react": "19.2.8",
     "react-dom": "19.2.8",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18,8 +18,8 @@ importers:
         specifier: 5.4.4
         version: 5.4.4
       electron-menubar:
-        specifier: 10.1.3
-        version: 10.1.3(electron@43.2.0(supports-color@10.2.2))
+        specifier: 10.1.4
+        version: 10.1.4(electron@43.2.0(supports-color@10.2.2))
       electron-updater:
         specifier: 6.8.9
         version: 6.8.9(supports-color@10.2.2)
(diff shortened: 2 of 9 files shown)


setchy:
should this logic be within electron-menubar based on this config param being set? https://github.com/gitify-app/gitify/blob/c83528dd93b0239aa26e9650547628ef4dc7c024/src/main/index.ts#L34

setchy:
an update - after updating to `v7.1.1`, the macOS dock icon was still present. restarting the app helped.  doesn't look like the root cause is resolved (yet)

PR #3116, opened 2026-07-29
perf(notifications): respect server-recommended poll interval

afonsojramos:
Fixes part of #3101

## Problem

Polling uses only the user-configured `fetchInterval` and ignores the `X-Poll-Interval` header GitHub returns on the notifications list endpoint. The header advertises the minimum poll spacing the server currently wants (normally `60`, raised when GitHub asks clients to back off), so Gitify keeps polling at the configured rate even when the server requests slower polling.

## Change

- New `pollInterval` registry (`src/renderer/utils/notifications/pollInterval.ts`): forge clients report the latest server-recommended interval per account, and `computeRefetchIntervalMs` returns the user interval stretched to the slowest recommendation across the currently polled accounts.
- The GitHub client reports `X-Poll-Interval` from both the single-page and paginated notifications list fetches.
- `useNotifications` passes `refetchInterval` as a function, so TanStack Query re-evaluates the interval after every fetch and picks up server changes without a query-key change.

Forges that do not send the header never report, and polling falls back to the user interval unchanged. With the default 60s fetch interval and GitHub's normal `X-Poll-Interval: 60` this is behavior-neutral; polling only slows when GitHub asks for more spacing (or when a user configures an interval below the server minimum).

## Tests

- `pollInterval.test.ts`: stretch-to-slowest semantics, never faster than the user interval, ignoring intervals from accounts no longer polled, ignoring invalid values.
- `client.test.ts`: header capture on both the single-page and paginated fetch paths.

### DIFF
--- a/src/renderer/utils/notifications/pollInterval.ts
+++ b/src/renderer/utils/notifications/pollInterval.ts
@@ -1,0 +1,61 @@
+import type { Account } from '../../types';
+
+/**
+ * Last server-recommended minimum poll interval in seconds, per account.
+ *
+ * GitHub returns an `X-Poll-Interval` header on the notifications list
+ * endpoint indicating how frequently clients may poll it (usually 60 seconds,
+ * raised when the server wants clients to back off). Forge clients report the
+ * header value here after each list fetch and the notifications query reads it
+ * back to stretch its polling interval accordingly.
+ *
+ * @see https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api
+ */
+const serverPollIntervalsSeconds = new Map<string, number>();
+
+/**
+ * Stable per-account key, derived from the same identity fields as
+ * `getAccountUUID`. Kept local so this leaf module does not import the
+ * auth/forge module graph (which would create an import cycle back through
+ * the forge clients); the key only needs to be consistent within this
+ * registry.
+ */
+function accountKey(account: Account): string {
+  return `${account.hostname}-${account.user?.id}-${account.method}`;
+}
+
+/**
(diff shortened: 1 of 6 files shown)


setchy (approved):

src/renderer/utils/notifications/pollInterval.ts
```diff
@@ -0,0 +1,61 @@
...
+function accountKey(account: Account): string {
+  return `${account.hostname}-${account.user?.id}-${account.method}`;
+}
+
+/**
+ * Record the server-recommended minimum poll interval for an account.
+ * Non-finite or non-positive values are ignored, so forges without the header
+ * simply never report.
+ */
+export function reportServerPollInterval(account: Account, seconds: number): void {
```
setchy, line 32: thinking out loud... how would this work between the REST api response and the GraphQL API response headers for a given account?  could those potentially return different x-poll intervals?  i had a feeling these API types have different quotas they draw from...

afonsojramos: X-Poll-Interval isn’t about quotas, it is purely a polling-cadence hint that only the notifications REST endpoint returns. 
Just verified live: GET /notifications sends `x-poll-interval: 60`, while other REST endpoints and GraphQL don’t send it at all (you’re right they draw from separate buckets: `x-ratelimit-resource: core` vs `graphql`). So there is no disagreement between the two. The registry is only written from the notifications list fetch, and since the GraphQL enrichment runs inside the same poll cycle, a raised interval slows both proportionally.

setchy (approved):

PR #3124, opened 2026-07-30
feat(ui): fade scrollable content at the top and bottom edges

afonsojramos:
Scrollable views currently cut off hard at the top and bottom edges, so content collides abruptly with the header and footer chrome. This adds a `scrollFade` opt-in to the shared `Contents` layout component that softly fades content in and out at both edges while scrolling.

The fade is driven entirely by CSS scroll-driven animations (`animation-timeline: scroll(self block)`) masking the scroll container, so there are no scroll listeners or re-renders, and each edge only fades while there is actually more content to scroll in that direction (measured in Chromium: no top fade at scroll position 0, no bottom fade at the end). Containers whose content fits leave the scroll timeline inactive and render no fade at all.

Enabled on Settings, Filters, Accounts, Scopes and the three login screens. Notifications is deliberately left out for now: read notifications are removed from the list, so it can go from overflowing to fitting while mounted, and Chromium holds the last fade value rather than clearing it, which would strand a fade above a list that can no longer scroll.

### DIFF
--- a/src/renderer/App.css
+++ b/src/renderer/App.css
@@ -32,6 +32,60 @@ body,
   background-color: var(--gitify-scrollbar-thumb-hover);
 }
 
+/**
+ * Softens the top and bottom edges of a scroll container, revealing each edge
+ * only while there is more content to scroll in that direction.
+ */
+@property --gitify-scroll-fade-top {
+  syntax: '<length>';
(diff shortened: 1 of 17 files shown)


setchy (approved):
❤️

PR #3130, opened 2026-07-31
feat(theme): add Glass design language

afonsojramos:
## Summary

After simmering for 2 days @setchy 👀 

Adds a second design language, **Glass**, next to the existing look (now Classic), selectable under Settings → Appearance. Under Glass the whole window becomes one continuous translucent material: the sidebar dissolves into icons floating on the glass, controls (buttons, selects, checkboxes, radios, count pills) render as stepped translucent tiles, notification and repo rows float as rounded cards with glass hover states, and tooltips/popovers come out frosted. Classic is pixel-unchanged: everything Glass lives behind `[data-theme='glass']` plus a `.gitify-translucent` gate, so there is no risk to the default experience.

image

image

image

## Design decisions

- **Native material where the OS offers one.** On macOS the window gets a real `popover` vibrancy material, wired over IPC and ordered so switching designs at runtime never flashes black; the material also follows the app's light/dark mode via `nativeTheme` sync. Other platforms fall back to CSS `backdrop-filter`.
- **One token layer, zero component branching.** Chrome values (radius, tints, stepped control fills, hairlines) are CSS custom properties re-pointed per language, and Primer's own design tokens are overridden at the `data-color-mode` boundary, so components stay stock Primer and the entire language lives in `App.css`.
- **Accessibility degradation.** OS Reduce Transparency / Increase Contrast switches Glass to solid, legible surfaces, and the vibrancy material is withheld entirely.
- **Native-feeling details.** Desaturated status palette, borderless frosted pills, translucent label tokens that keep their GitHub hue, slim overlay scrollbar, and hover actions that fade row text into the glass instead of painting an opaque band.

## Validation

- Full test suite: 160 files / 1245 tests passing; `tsc`, lint and format clean.
- Exercised live on macOS in light and dark: notifications list, settings, filters; hover, focus, disabled and reduced-transparency states; runtime Classic ↔ Glass switching.

### Theme tokens that re-resolve per language (Tailwind `@theme inline`)

The radius/shadow/tint utilities are backed by plain CSS variables (`--gitify-*`) that each design language re-points under its own `[data-theme]` scope. Declaring them with `@theme inline` (instead of `@theme`) makes Tailwind re-resolve the utilities against the *active* scope rather than freezing them at `:root`.

```css
@theme inline {
  --radius-gitify-md: var(--gitify-radius-md); /* follows the active [data-theme] */
}
```

This is why one component tree can serve two languages with no JS branching: the swap is a single attribute on the root. Reach for it when a whole token layer swaps at once; for one-off per-component tweaks, a direct class is simpler.

---

Note: I now fully prefer this theme 👀


setchy (changes requested):
Marking while I have time to review.

src/main/handlers/system.ts
```diff
@@ -84,4 +91,24 @@ export function registerSystemHandlers(mb: Menubar): void {
...
+   * Request/response so the renderer can await the material before clearing the
+   * window's own background (avoids a black frame during the switch).
+   */
+  handleMainEvent(EVENTS.SET_WINDOW_VIBRANCY, (_, enabled: boolean) => {
+    applyWindowVibrancy(mb, enabled);
+    return undefined;
+  });
+
+  /**
+   * Sync the native appearance with the app's color mode so the macOS vibrancy
```
setchy, line 106: i wasn't able to find where in macOS System Settings the vibrancy setting is.  Can you point me in the right direction :)

src/renderer/__mocks__/state-mocks.ts
```diff
@@ -24,8 +25,8 @@ export const mockAuth: { accounts: Account[] } = {
 export const mockToken = 'token-123-456' as Token;
 
 const mockAppearanceSettings: AppearanceSettingsState = {
+  designLanguage: DesignLanguage.CLASSIC,
   theme: Theme.SYSTEM,
-  increaseContrast: false,
```
setchy, line 28: this was used to control an accessibility feature of the GitHub primer themes

src/renderer/components/settings/AppearanceSettings.tsx
```diff
@@ -90,7 +102,14 @@ export const AppearanceSettings: FC = () => {
               unsafeDisableTooltip={true}
             />
 
-            <Button aria-label="Zoom percentage" disabled size="small">
+            {/* Value display, not a control: styled like the group but inert
```
setchy: great fix!

setchy: we repeat this a few times - should we extract it out?  or ok to duplicate?

src/renderer/__mocks__/state-mocks.ts
setchy: Image

setchy: was your thoughts to control this from the macOS Accessibility `increase contrast` setting directly?

setchy:
Super sharp enhancement, @afonsojramos.  I love that we're able to keep the design components as Primer and apply new theme/style overrides.  Nice work.  

Left a few inline comments for my own curiosity

src/renderer/hooks/useAppearance.ts
```diff
@@ -0,0 +1,88 @@
...
+      window.gitify.platform.isMacOS() ? 'vibrancy' : 'backdrop-filter',
+    );
+  }, []);
+
+  useEffect(() => {
+    const root = document.documentElement;
+
+    // Glass is always translucent; it only degrades to solid under the OS Reduce
+    // Transparency / Increase Contrast settings (a media query in App.css).
+    const vibrant = designLanguage === DesignLanguage.GLASS && !prefersReducedTransparency;
```
setchy: it might be nice to allow users to set `prefersReducedTransparency` just within Gitify, as well as sync it from the macOS system settings.

src/main/handlers/system.ts
afonsojramos: There isn’t a “vibrancy” toggle as such. Vibrancy is the native translucency material (NSVisualEffectView) that macOS uses for menus, sidebars, the menu bar and so on. We just ask the OS for it via the window’s vibrancy option, which is what gives Glass its frosted look. The one user-facing control that affects it is System Settings → Accessibility → Display → “Reduce transparency”: switch that on and macOS renders these materials solid, which we honor (the window falls back to a solid surface via prefers-reduced-transparency).

src/renderer/hooks/useAppearance.ts
afonsojramos: That's fair. Again, as said in the other comment, as is, we read it from the OS, but we can also read it from an in-app setting, but I'd say that it is not common to see this as an app-specific option. The whole point of this theme is the native view. Maybe adjusting the transparency might be a better fit, but I'd argue that that is separate work to this pr, as a follow-up.

src/renderer/components/settings/AppearanceSettings.tsx
afonsojramos: I mean... There's only three places, I think it is fine to leave it. I can remove the comments tho.

src/renderer/__mocks__/state-mocks.ts
afonsojramos: Whoops, no, I just meant to remove it from Glass, but forgot to add it back in Classic. Re-adding.

src/renderer/hooks/useAppearance.ts
setchy: sharing an example from another taskbar app - [OpenUsage](https://www.openusage.ai/) have this as a settings toggle

Screenshot 2026-08-02 at 8 52 09 AM
Screenshot 2026-08-02 at 8 52 15 AM

setchy:
Was doing more manual testing this morning - the notification type icon colors appear broken [always black]

setchy:
When in glass mode, the checkbox for `Show full notification title` is missing some styles
2026-08-04T15-09-06 565Z-Gitify-screenshot

you can still toggle it, but when unchecked its missing the checkbox outline

src/renderer/__mocks__/state-mocks.ts
setchy: @afonsojramos - `increaseContrast` is missing from our `mockAppearanceSettings`

afonsojramos:
Are you sure that you ran everything correctly? This is what it should look like
image

src/renderer/__mocks__/state-mocks.ts
afonsojramos: It’s back already?? `increaseContrast` was restored in bba2ca87 along with the Classic-only checkbox and the new prefers-contrast: more wiring, so mockAppearanceSettings has it at state-mocks.ts:30

setchy:
> Are you sure that you ran everything correctly? 

Yup, fairly certain - try toggling the theme - looks like several other components like the footer buttons also loose their styling when set to `Glass` + `System`


https://github.com/user-attachments/assets/f848af9f-4e9d-49ee-8f2e-607d20f239a9

setchy:
> Was doing more manual testing this morning - the notification type icon colors appear broken [always black]

| branch | screenshot |
| --- | --- |
| **main** | Screenshot 2026-08-04 at 5 08 17 PM |
| **native-theme-support** - Classic | Screenshot 2026-08-04 at 5 10 27 PM |
| **native-theme-support** - Glass | Screenshot 2026-08-04 at 5 09 30 PM |

setchy (approved):
lgtm 🚀  (i pushed some small changes to address the sonarqube issues)

curious if we would want colored-icons when in Glass design mode.

PR #3147, opened 2026-08-04
fix: correct online state on startup instead of waiting for a network event

setchy:
**Problem**
TanStack Query's onlineManager initializes to online: true regardless of the device's actual network state, and only self-corrects when the browser fires a native online/offline event. Since browsers only emit events on transitions, a device that's already offline when Gitify launches (airplane mode, VPN down, etc.) never gets one — so the app incorrectly believes it's online from the first render, fires the notifications query, and only discovers it's offline after that query fails and retries.

**Fix**
Added a single onlineManager.setOnline(navigator.onLine) call at the start of useOnlineStatus's mount effect, before the existing subscription logic. This actively force-corrects onlineManager's state to match reality on startup, instead of passively waiting for a future browser event that may never come.

### DIFF
--- a/src/renderer/hooks/useOnlineStatus.test.ts
+++ b/src/renderer/hooks/useOnlineStatus.test.ts
@@ -12,6 +12,7 @@ describe('renderer/hooks/useOnlineStatus.ts', () => {
     act(() => {
       onlineManager.setOnline(true);
     });
+    Object.defineProperty(navigator, 'onLine', { value: true, configurable: true });
   });
 
   it('reflects the online manager status', () => {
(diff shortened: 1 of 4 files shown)


afonsojramos (approved):
Verified the offline cold-start path and the follow-ups pushed here.

Moving the correction to `queryClient.ts` module scope means it runs at import, before the client exists and therefore before any query can fire — so it no longer depends on `GlobalEffects` rendering ahead of the router or on `useOnlineStatus()` being declared above `useNotifications()`. Seeding `useState` from `onlineManager.isOnline()` keeps a single source of truth and drops the one frame that painted as online.

Also replaced the `confirms onlineManager as online when the device starts online` case: it passed with the production line deleted (onlineManager already defaults to `true`, and `setOnline` only notifies on change). The replacements start from the opposite state in both directions and I confirmed both fail without the fix.

PR #3178, opened 2026-08-13
fix(forge/github): sanitize gql query docs based on gated feature compatibility

### LINKED ISSUE #3174: GitHub Enterprise: PR enrichment fails because PullRequest.stackEntry is queried unconditionally in 7.3.x

### 📝 Description

It looks like starting from 7.3.0, enrichGitHubNotifications fails with the following error: `GraphQL request returned errors: Field 'stackEntry' doesn't exist on type 'PullRequest'`

This is causing the status icons to show in grey (inspecting in DevTools indicates that they're labeled with `text-gitify-icon-muted`)

AFAIK though there isn't a known upcoming version of GHE that will include stacked PRs but hopefully soon

### 🪜 Steps To Reproduce

1. Connect gitify to GHE instance
2. Open DevTools
3. Inspect console when retrieving notifications

### 🪵 Log Excerpts

_No response_

### Gitify Version

7.3.2

### Operating System

macOS

### GitHub Account

GitHub Enterprise Server

### 📸 Screenshots

_No response_

setchy:
Fixes: #3174 (attempt two 😅)

GHES doesn't expose PullRequest.stackEntry (or Discussion.isAnswered before 3.12). 

The old `@include(if:)` guard didn't help, since GraphQL Servers will validate the query document field names regardless of client-side directives, and rejected the unused variable, breaking PR enrichment for all GHE users.

This adds a custom `@gated(requires: "...")` directive, marks the two fields with it, and strips unsupported fields (plus the directive) from the query before sending via stripGatedSelections(). 

`@include/@skip` only affect execution, so physically removing the selection is the only reliable fix; the directive keeps it declarative in the sources and codegen still types the fields.

### DIFF
--- a/src/renderer/utils/forges/github/capabilities.test.ts
+++ b/src/renderer/utils/forges/github/capabilities.test.ts
@@ -5,6 +5,7 @@ import {
 
 import {
   githubCapabilities,
+  getGitHubCapabilities,
   supportsAnsweredDiscussion,
   supportsStackedPullRequests,
 } from './capabilities';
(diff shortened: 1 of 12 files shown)


setchy:
@sphinxrave - if you have a moment, we would appreciate if you could test the build from this branch. Should resolve the GHE stacked issue

PR #3200, opened 2026-08-16
chore: init openspec

setchy:
https://openspec.dev/


afonsojramos:
Just noticed that this got added to the repo without much fuss. What was the reasoning here? Can't it stay as a local machine thing?

PR #3201, opened 2026-08-16
feat(updater): add setting to opt-out of update notifications

### LINKED ISSUE #3181: Can update notifications be disabled?

### 📝 Description

I've seen a few closed issues for this without resolution; is there a way to disable update notifications? I installed gitify through Homebrew and I manage updates that way - but gitify notifies me a few times a day that an update is available.

### 🪜 Steps To Reproduce

1. Install gitify via homebrew
2. Wait for a new version to come out
3. Get multiple update notifications throughout the day

### 🪵 Log Excerpts

_No response_

### Gitify Version

7.3.1

### Operating System

macOS

### GitHub Account

GitHub Cloud

### 📸 Screenshots

_No response_

nicholas-ochoa:
Add a new setting that allows users to disable automatic updates

Fixes #3181


setchy (changes requested):
Thanks for contributing. 

I'd like to explore, if possible, how we could better detect if it's installed via homebrew (i believe they already have this capability).

I'd really like to keep users as close to the latest mainline release. Is there something better we could offer than disabling auto updates completely?

nicholas-ochoa:
Detecting if gitify was installed via Homebrew is possible. How do you feel about only showing the checkbox when a package manager install is detected?

I'm not sure about what other approaches can be taken, I'm open to any ideas on a different direction to take with this PR.

I'm just trying to solve my original issue with the frequent notifications popping up that are not needed (since Homebrew is updating it, in my case).

afonsojramos:
I'd say that no checkbox needs to exist. We can just disable it when we detect that it was indeed installed via a package manager. Right?

nicholas-ochoa:
That approach would work for me

nicholas-ochoa:
I've reworked the PR a bit. It now detects if Gitify was installed via Homebrew (it should be flexible enough to add support for other package managers on Linux pretty easily). If a package manager install is detected, the automatic update check / notification is disabled and the tray menu just shows a disabled "Updates managed by Homebrew" entry so users aren't confused on why the menu option isn't displayed.

afonsojramos:
So, I was looking into this and I've detected something that I think we missed in this discussion. Which is that in 

https://github.com/Homebrew/homebrew-cask/blob/532d42db709fcff52cb213407a9c3d533ab5fdad/Casks/g/gitify.rb#L15

We have the following `auto_updates true`, which actually tells homebrew that the app updates itself. So, if we are to move forward with this PR, we need to remove that entry too. To be honest, I think it is a good direction for us to move towards, what do you think @setchy?

setchy:
> So, I was looking into this and I've detected something that I think we missed in this discussion. Which is that in 
> 
> https://github.com/Homebrew/homebrew-cask/blob/532d42db709fcff52cb213407a9c3d533ab5fdad/Casks/g/gitify.rb#L15
> 
> We have the following `auto_updates true`, which actually tells homebrew that the app updates itself. So, if we are to move forward with this PR, we need to remove that entry too. To be honest, I think it is a good direction for us to move towards, what do you think @setchy?

Ah, great find - I hadn't thought to check the brew config.

What if as an alternative approach we
- kept the brew cask as-is (app updated)
- kept the auto update schedule as-is
- added a System Setting for users to opt-out of Update system notifications. 

This way we're 
- less noisy to end users
- still controlling the update cadence. 
- consistent for all OS platforms

The system menu will already hint when an update is ready to install on next restart (system or app launch). 

We simply allow user preference of the update available notification.

nicholas-ochoa:
I'm happy with that option - the notifications are the annoying part, I don't have an issue with the auto update itself.

afonsojramos:
Yeah, however, I'd say that a user who installs via Homebrew generally prefers for the update to be handled via Homebrew, don't you think? Otherwise, they would have installed via `.dmg`, no?

setchy:
How about for this PR, we keep it focused on a new user-setting to opt-out of auto-update notifications (the update schedule should still check and report via menubar).

In a separate GH Issue, we could explore how our auto-update process interacts with package manager installations [homebrew, nixpkgs, etc]

setchy (approved):

PR #3208, opened 2026-08-20
fix(deps-core): pin electron to 43.2.0 to restore Linux tray icons

### LINKED ISSUE #3206: Icon does not appear in systray

### 📝 Description

Hi!

I was way behind and I decided to upgrade recently and the systray was gone. I was able to target 7.3.0 as the version that introduced the problem. I'm using the "AppIndicator and KStatusNotifierItem Support" GNOME extensions on Fedora 43 (wayland).

Essentially a three-dots icon appears when the program launches but then it diseapears while Gitify still runs in the background. I can launch it using `$ ./Gitify.AppImage --inspect=9229`, inspect it and use `require('electron').BrowserWindow.getAllWindows().forEach(w => { w.show(); w.focus(); });` to show its main window.

### 🪜 Steps To Reproduce

Launch Gitify

### 🪵 Log Excerpts

```console
Debugger listening on ws://127.0.0.1:9229/918dcfb0-dcbe-42ac-becb-07736b04f98a
For help, see: https://nodejs.org/learn/getting-started/debugging
15:54:05.117 › [app updater] Starting updater
15:54:05.121 › [app updater] Checking for updates on application launch
15:54:05.122 › [auto updater] Checking for update
[330463:0818/155405.513338:ERROR:components/dbus/xdg/request.cc:173] Request ended (non-user cancelled).
[330463:0818/155405.513389:ERROR:ui/base/accelerators/global_accelerator_listener/global_accelerator_listener_linux.cc:358] Failed to call BindShortcuts (error code 5).
15:54:07.370 › [auto updater] Update not available
Debugger attached.
```

### Gitify Version

7.3.0 and 7.4.0

### Operating System

Linux

### GitHub Account

GitHub Cloud

### 📸 Screenshots

_No response_

afonsojramos:
## Summary

Pins `electron` back to **43.2.0** (from 43.4.0) and adds a Renovate hold so the broken range can't come back automatically.

Electron 43.3.0 pulled in a Chromium change ([electron#52416](https://github.com/electron/electron/pull/52416), Chromium 150.0.7871.181) that replaced per-icon D-Bus object paths with a *StatusNotifierItem multiplexer*. The multiplexer collapses every tray icon onto a single `/StatusNotifierItem` path, leaving the destination service name as the only discriminator. Desktops that address the item by its **unique** D-Bus connection name rather than its well-known name (anything built on `Gio.DBusProxy`) can no longer be routed to the right icon, so `Multiplexer::GetIcon(":1.x")` finds nothing and **no tray icon renders**.

Gitify is a tray-first app, so on affected desktops this makes it effectively unusable.

## Confirmed in the wild

#3206 is this bug, reported by a user running the **packaged AppImage** on Fedora 43 / GNOME (Wayland) with the AppIndicator extension. They independently bisected to v7.3.0, which lines up exactly with the Electron bump:

| Gitify | Electron | Tray |
| --- | --- | --- |
| v7.2.0 | 43.2.0 | works |
| v7.3.0 | 43.3.0 | **broken** |
| v7.3.1 / v7.3.3 | 43.3.0 | **broken** |
| v7.4.0 | 43.4.0 | **broken** |

Their symptom, a three-dots placeholder that appears at launch then vanishes, matches the upstream reporter in electron#52674 verbatim. This also settles a question that was otherwise open: packaged electron-builder builds are affected, not only directly-launched Electron.

Both currently-supported release lines (7.3.x and 7.4.x) are therefore shipping a broken tray on affected desktops, so a patch release should follow this merge.

## Evidence

Reproduced in [electron-menubar](https://github.com/gitify-app/electron-menubar)'s visual tray matrix, which boots a real menubar fixture on 14 platforms and asserts the tray icon and popover are actually painted. Three runs the same day, on the same runner image and harness, differing only in Electron version:

| Electron | Budgie | Sway/Waybar | LXQt | KDE / macOS / Windows |
| --- | --- | --- | --- | --- |
| 43.2.0 | pass | pass | pass | pass |
| 43.4.0 | **fail** | **fail** | pass | pass |
| 43.4.1 | **fail** | **fail** | **fail** | pass |

In every failing case the popover window renders with exactly the expected pixel counts and only the tray icon is missing, so this is specific to tray rendering. The harness re-captures and re-checks for 60s before failing, so a slow-painting panel is ruled out. Failures reproduced on re-run.

**43.4.1 does not fix this**, despite a release note that reads like it does. That change ([electron#52952](https://github.com/electron/electron/pull/52952)) is described upstream as a bad fix for the same problem; we measured it removing waybar's `Invalid Status Notifier Item` error without making the icon appear, and LXQt newly broke on it. That is why the hold is `<43.3.0` rather than "skip 43.4.0".

Upstream tracking: [electron#52674](https://github.com/electron/electron/issues/52674) (open, no fix PR yet). [electron#53024](https://github.com/electron/electron/issues/53024) was closed as a duplicate.

## Changes

- `package.json` and `pnpm-workspace.yaml` — Electron is pinned in both the `devDependencies` entry and the `catalog` that feeds `overrides.electron: 'catalog:'`, so both move together.
- `renovate.json` — adds an `allowedVersions: "<43.3.0"` hold, following the existing `got` pattern, with a description recording the removal criteria: electron#52674 closed **and** electron-menubar's visual matrix green on the candidate.

## Verification

- `pnpm run build` and `pnpm run check` (format + lint, 442 and 392 files) pass.
- Full CI is green, including all three `electron-builder` build jobs (Linux, Windows, macOS) and the test suite.
- Locally `pnpm test` showed 3 failures in `LoginWithOAuthApp.test.tsx` and `LoginWithPersonalAccessToken.test.tsx`; these are pre-existing load-induced timeouts (both files pass 19/19 in isolation, neither touches Electron) and they pass in CI.
- Lockfile diff is electron-only.

Closes #3206

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "concurrently": "10.0.5",
     "date-fns": "4.4.0",
     "dotenv": "17.4.2",
-    "electron": "43.4.0",
+    "electron": "43.2.0",
     "electron-builder": "26.15.3",
     "final-form": "5.0.1",
     "graphql": "17.0.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,7 +8,7 @@ overrides:
   react-is: 19.2.8
   cross-spawn: 7.0.6
   got: 11.8.6
-  electron: 43.4.0
+  electron: 43.2.0
 
 importers:
 
@@ -19,7 +19,7 @@ importers:
         version: 5.4.4
       electron-menubar:
         specifier: 10.2.0
-        version: 10.2.0(electron@43.4.0(supports-color@10.2.2))
+        version: 10.2.0(electron@43.2.0(supports-color@10.2.2))
       electron-updater:
         specifier: 6.8.9
         version: 6.8.9(supports-color@10.2.2)
@@ -145,8 +145,8 @@ importers:
         specifier: 17.4.2
         version: 17.4.2
       electron:
-        specifier: 43.4.0
-        version: 43.4.0(supports-color@10.2.2)
+        specifier: 43.2.0
+        version: 43.2.0(supports-color@10.2.2)
       electron-builder:
         specifier: 26.15.3
         version: 26.15.3(electron-builder-squirrel-windows@26.15.3)(supports-color@10.2.2)
@@ -2551,7 +2551,7 @@ packages:
   electron-menubar@10.2.0:
     resolution: {integrity: sha512-CMhVE91BZg5OQkViOHOknM2+HQMwwopxJjNR+Uuu7flc37I/MdujpeOgGPTAQJXgPnwYsPZ+VXo5K2kB+QrLpA==}
     peerDependencies:
-      electron: 43.4.0
+      electron: 43.2.0
 
   electron-publish@26.15.3:
     resolution: {integrity: sha512-g/2bn8YTavY4cuS5F+jOS7zmZbXXBV8KZ8yHKfJjFPoKtzBqrpCdNPxBd3tqdBwP7BVd0lGzf7Bk2s0KesWZ4Q==}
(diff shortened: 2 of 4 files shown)


setchy:
Great detective work, @afonsojramos! 🕵️

afonsojramos:
@setchy the visual tests in electron-menubar are worth it after all! 😅

PR #3225, opened 2026-08-26
feat(forge): add GitLab support

### LINKED ISSUE #237: New Forge: GitLab support
Would be nice with Gitlab support as well!

afonsojramos:
Adds GitLab (Cloud and Self-Managed) as a forge behind the existing `ForgeAdapter` interface, with no forge-specific branching outside `src/renderer/utils/forges/gitlab/`. GitLab has no `/notifications` endpoint, so the adapter is built on the To-Do API — and because a to-do embeds its full target object, `enrichNotifications` is omitted entirely and one list call carries everything the detailed-notifications setting needs. GitLab also has no read/unread axis (only `pending`/`done`), so `capabilities.markAsDone` reports `false` and mark-as-read maps onto `POST /todos/:id/mark_as_done`, keeping the row to a single button.

Worth flagging for review: clearing a to-do needs the broad `api` scope on legacy tokens (or the `Todo` resource on GitLab 18+ fine-grained tokens), and To-Do items are narrower than GitHub notifications — both are stated on the login screen. Every API behaviour was verified against live gitlab.com rather than taken from the docs, and the feature was exercised end-to-end in the app: login, render, mark-as-read round-trip, and logout.

Closes #237


setchy:
Great work @afonsojramos - our forges continue to go from strength to strength.

I don't have a GitLab handy to test on, but lgtm

setchy (approved):

PR #3245, opened 2026-09-01
ci: use setup-vp for JavaScript tooling

afonsojramos:
## Summary

CI now lets the official Vite+ action own Vite+, Node.js, pnpm, dependency installation, and pnpm-store caching instead of composing those pieces ourselves. This isolates whether Vite+ materially simplifies Gitify's workflow setup before we decide on the larger toolchain direction.

`setup-vp` resolves the existing project pins: Vite+ 0.2.9 from `package.json`, Node.js 24.20.0, and pnpm 11.24.0 from `packageManager`. Jobs that need dependencies use the action's default `vp install`; Renovate config validation keeps installation disabled.

The action uses the pinned package manager internally but does not expose a bare `pnpm` executable to later workflow steps. Those steps now use Vite+'s supported command surface instead: `vp run` for package scripts, `vp exec` for project binaries, and `vp dlx` for the Renovate validator.

Related: #3244

## Validation

- This PR's CI passed environment detection, cache restoration, installation, lint, TypeScript, unit tests, visual regression, SonarQube analysis, and packaging on Linux, macOS, and Windows.
- The repository's pinned actionlint 1.7.12 accepts every workflow.
- Renovate config validation passes through `vp dlx --package renovate renovate-config-validator`.
- zizmor 1.30.0 reports no findings from the new external action usage; its remaining findings are pre-existing repository advisories.


afonsojramos:
@setchy I think that the issue might have been that we only partially adopted vp. With full adoption, we might have no more issues.

setchy (approved):

setchy:
Good catch, this is much more clean 🧼✨.

Looks like we still have the same errors on our two PRs (one legit lint issue, one a version mismatch)

afonsojramos:
> Looks like we still have the same errors on our two PRs (one legit lint issue, one a version mismatch)

Addressed in #3249: upgraded Vite+ to 0.3.0 with both direct Vitest providers at 4.1.11, fixed the two new React lint diagnostics, and grouped the three coupled packages in Renovate so future updates land together.

PR #3255, opened 2026-09-03
refactor(shortcuts): forge specific issue and pull request urls

### LINKED ISSUE #3253: Gitlab: incorrect URLs for issues and merge requests shortcut buttons

### 📝 Description

The two buttons on the right redirect to Github formatted URLs:
- `<gitlab_url>/pulls` instead of `<gitlab_url>/dashboard/merge_requests`
- `<gitlab_url>/issues` instead of `<gitlab_url>/dashboard/work_items`

Image

### 🪜 Steps To Reproduce

Click on the shortcuts on a Gitlab account

### 🪵 Log Excerpts

_No response_

### Gitify Version

7.7.0

### Operating System

macOS

### GitHub Account

GitHub Cloud

### 📸 Screenshots

_No response_

setchy:
closes #3253


src/renderer/utils/forges/gitea/adapter.ts
```diff
@@ -96,6 +96,9 @@ export const giteaAdapter: ForgeAdapter = {
     `https://${hostname}/user/settings/applications` as Link,
   getAccountSettingsUrl: (account: Account) =>
     `https://${account.hostname}/user/settings/applications` as Link,
+  getIssuesUrl: (account) => `https://${account.hostname}/issues` as Link,
+  getPullRequestsUrl: (account) => `https://${account.hostname}/pull` as Link,
```
afonsojramos: Gitea's route is `/pulls`, not `/pull`. Both `https://gitea.com/pull` and `https://codeberg.org/pull` return 404, while `/pulls` redirects to login on both.

The adapter test and `links.test.ts` lock in the wrong path, so they need updating too.

src/renderer/utils/forges/gitlab/adapter.ts
```diff
@@ -168,6 +168,11 @@ export const gitlabAdapter: ForgeAdapter = {
     `https://${hostname}/-/user_settings/personal_access_tokens` as Link,
   getAccountSettingsUrl: (account: Account) =>
     `https://${account.hostname}/-/user_settings/personal_access_tokens` as Link,
+  getIssuesUrl: (account) => `https://${account.hostname}/dashboard/work_items` as Link,
```
afonsojramos: `/dashboard/work_items` only exists from GitLab 18.10 (absent from `config/routes/dashboard.rb` through 18.9), so this 404s on older self-managed instances. `/dashboard/issues` exists on every version and, from 18.10, 301-redirects to the work items dashboard when the consolidated list is enabled for the user.

GitLab's own "Your work" sidebar also links with `?assignee_username=<login>`, and `DashboardController` runs `check_filters_presence!` on both `issues` and `work_items`, so a bare URL may land on the "select at least one filter" empty state.

Suggest `/dashboard/issues?assignee_username=${account.user.login}` (the profile link in `links.ts` already relies on `account.user` being set).

`/dashboard/merge_requests` and `/dashboard/todos` match GitLab's own nav and look right.

src/renderer/hooks/useShortcutActions.ts
```diff
@@ -60,7 +60,7 @@ export function useShortcutActions(): { shortcuts: ShortcutConfigs } {
       myNotifications: {
         key: 'n',
         isAllowed: isLoggedIn,
-        action: () => openHostNotifications(primaryAccountHostname),
+        action: () => primaryAccount && openHostNotifications(primaryAccount),
```
afonsojramos: Behaviour change worth a deliberate decision: `Sidebar` renders the notifications, issues and pull request buttons even when logged out, and on `main` they fell back to github.com via `primaryAccountHostname`. With this guard they become silent no-ops when there is no account, which is what the three Sidebar tests caught before they were updated to seed an account.

Either keep a GitHub fallback when there is no account, or hide those buttons when logged out. Swallowing the click seems like the worst of the three options. Same applies to `myIssues` and `myPullRequests` below.

src/renderer/utils/forges/gitea/adapter.ts
afonsojramos: Gitea's route is `/pulls`, not `/pull`. Both `https://gitea.com/pull` and `https://codeberg.org/pull` return 404, while `/pulls` redirects to login on both.

```suggestion
  getPullRequestsUrl: (account) => `https://${account.hostname}/pulls` as Link,
```

The adapter test and `links.test.ts` lock in the wrong path, so they need the same change.

src/renderer/utils/forges/gitea/adapter.ts
```diff
@@ -96,6 +96,9 @@ export const giteaAdapter: ForgeAdapter = {
...
+  getNotificationsUrl: (account) => `https://${account.hostname}` as Link,
```
afonsojramos: This is a regression: `main` opened `/notifications` for every host, and that is a real Gitea page (served by gitea.com and codeberg.org).

```suggestion
  getNotificationsUrl: (account) => `https://${account.hostname}/notifications` as Link,
```

The adapter test and `links.test.ts` expect the host root, so they need updating too.

src/renderer/utils/forges/gitlab/adapter.ts
afonsojramos: `/dashboard/work_items` only exists from GitLab 18.10 (absent from `config/routes/dashboard.rb` through 18.9), so this 404s on older self-managed instances. `/dashboard/issues` exists on every version and, from 18.10, 301-redirects to the work items dashboard when the consolidated list is enabled for the user.

GitLab's own "Your work" sidebar also links with `?assignee_username=<login>`, and `DashboardController` runs `check_filters_presence!` on both `issues` and `work_items`, so a bare URL may land on the "select at least one filter" empty state.

```suggestion
  getIssuesUrl: (account) => {
    const url = new URL(`https://${account.hostname}/dashboard/issues`);
    if (account.user) {
      url.searchParams.set('assignee_username', account.user.login);
    }
    return url.toString() as Link;
  },
```

The adapter test and `links.test.ts` will need the matching expectation (the mock user supplies the login). `/dashboard/merge_requests` and `/dashboard/todos` match GitLab's own nav and look right.

src/renderer/hooks/useShortcutActions.ts
afonsojramos, line 63: Behaviour change worth a deliberate decision: `Sidebar` renders the notifications, issues and pull request buttons even when logged out, and on `main` they fell back to github.com via `primaryAccountHostname`. With this guard they become silent no-ops when there is no account, which is what the three Sidebar tests caught before they were updated to seed an account.

Either keep a GitHub fallback when there is no account, or hide those buttons when logged out. Swallowing the click seems like the worst of the three options. Same applies to `myIssues` and `myPullRequests` below.

src/renderer/stores/useAccountsStore.ts
```diff
@@ -164,6 +164,10 @@ const useAccountsStore = create<AccountsStore>()(
         return get().accounts[0]?.hostname ?? Constants.GITHUB_HOSTNAME;
       },
 
+      primaryAccount: () => {
```
afonsojramos, line 163: With this in place `primaryAccountHostname` is only referenced by its own tests. Since this PR is what orphaned it, worth removing it here along with its tests and the `AccountsActions` entry.

src/renderer/utils/forges/bitbucket/adapter.ts
```diff
@@ -123,6 +123,12 @@ export const bitbucketAdapter: ForgeAdapter = {
 
   getAccountSettingsUrl: (_account: Account) => ATLASSIAN_TOKEN_SETTINGS_URL,
 
+  // Bitbucket does not capture a workspace slug, so shortcut links target the
+  // host root rather than a workspace-scoped ("/workspace/{slug}/...") URL.
+  getIssuesUrl: (account) => `https://${account.hostname}` as Link,
+  getPullRequestsUrl: (account) => `https://${account.hostname}` as Link,
```
afonsojramos: Optional: Bitbucket's "Your work" page at `/dashboard/overview` lists the pull requests you authored and are reviewing, and it exists. It would be a more useful target than the host root I think.

```suggestion
  getPullRequestsUrl: (account) => `https://${account.hostname}/dashboard/overview` as Link,
```

If you take this, the comment above and the adapter and `links.test.ts` expectations need adjusting.

afonsojramos (changes requested):
Niceee! I am loving the modularity that we are achieving. Have dropped a couple of URL fixes.

setchy:
> Niceee! I am loving the modularity that we are achieving. Have dropped a couple of URL fixes.

it really is nice to have this flexibility.

---

One pattern that felt a little smelly in this PR was the double-passing of `account`

eg:

```
getAdapter(account).getIssuesUrl(account)
```

something to perhaps explore later on...

afonsojramos (approved):

PR #3266, opened 2026-09-03
refactor(forges): bind account-scoped adapter operations via getAccountAdapter

afonsojramos:
## Summary

- Split the forge adapter contract. `ForgeAdapter` keeps the forge-wide members (login flows, token validation, display helpers), and a new `ForgeAccountAdapter` carries the account-scoped operations with no account parameter.
- Adapters implement those operations under `accountOps`, still taking the account explicitly; the shape is derived from `ForgeAccountAdapter` by a mapped `WithAccount` type so the two cannot drift. `getAccountAdapter(account)` in the registry binds them, so call sites read `getAccountAdapter(account).listNotifications()` instead of `getAdapter(account).listNotifications(account)`.
- Migrates all 18 call sites, re-points test spies at `adapter.accountOps.*`, and adds registry tests for the binding (argument forwarding, nested capability bundles, optional bundles, call-time member lookup).

Follows up on the double-passing pattern raised in https://github.com/gitify-app/gitify/pull/3255#issuecomment-5529692319.

## Notes

- No behaviour change: URLs, capability checks and notification flows are untouched. The full unit suite, typecheck, lint and format pass.
- A bound view closes over the account object it was created from, and the store replaces accounts on token rotation, so views should be created at the call site rather than stored. This is documented on `ForgeAccountAdapter`.
- `followUrl<T>` stays generic on the bound side; on the implementation side the type parameter erases to `unknown`, matching the trust boundary that existed before.

### DIFF
--- a/src/renderer/components/notifications/NotificationFooter.tsx
+++ b/src/renderer/components/notifications/NotificationFooter.tsx
@@ -6,7 +6,7 @@ import { cn } from 'cn';
 
 import { type GitifyNotification, Opacity, Size } from '../../types';
 
-import { getAdapter } from '../../utils/forges/registry';
+import { forAccount } from '../../utils/forges/registry';
 import { openUserProfile } from '../../utils/system/links';
 import { AvatarWithFallback } from '../avatars/AvatarWithFallback';
 import { MetricGroup } from '../metrics/MetricGroup';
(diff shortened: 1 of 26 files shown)


afonsojramos:
What do you think of these ergonomics @setchy? I'm a bit on the fence with the naming. Chained, `forAccount` makes sense, but it isolation it is awful. `getAccountAdapter` is much cleaner, but it does use a lot more characters.

setchy:
> What do you think of these ergonomics @setchy? I'm a bit on the fence with the naming. Chained, `forAccount` makes sense, but it isolation it is awful. `getAccountAdapter` is much cleaner, but it does use a lot more characters.

let me soak on this PR for a moment 🧽

afonsojramos:
@setchy just replaced the recursive binder and mapped types with explicit typed bindings. The account-bound API now stays the same, and followUrl<T> preserves its generic contract throughout.

setchy (approved):
Let's give it a go.

PR #3288, opened 2026-09-07
fix: Rotate icon, not whole button

SiriusCrain:


setchy:
Thanks for raising, @SiriusCrain.

I've fixed this by properly importing `@primer` motion classes - see https://github.com/gitify-app/gitify/pull/3289

PR #3316, opened 2026-09-16
feat: sign in with an existing GitHub CLI login

fmguerreiro:
Some orgs forbid classic PATs and will not approve third-party OAuth apps, leaving Gitify no working sign-in. The GitHub CLI's OAuth app is often already authorized there, so this reuses the token `gh auth token` holds — including one `gh` takes from `GH_TOKEN`, which for some users is its only credential.

The CLI stays the owner: every Octokit request resolves the token from `gh`, and re-reads it once on a 401, so rotation by `gh auth login` or `gh auth refresh` needs no re-login. The account row persists no credential.

Gitify cannot widen the scopes `gh` issues, so these accounts are judged on notification access (`notifications` or `repo`), not PAT scope names. The scopes page cannot widen them either, so it points at `gh auth refresh -s notifications`.

| Login screen | CLI sign-in |
| --- | --- |
| new GitHub CLI entry | hostname only, no token field |
| login | cli |


fmguerreiro:
rebased this down to just the cli login. the two perf commits moved out to #3319 and #3320.

PR #3322, opened 2026-09-18
fix: show menubar popup over fullscreen Spaces on macOS

### LINKED ISSUE #563: Gitify jumps to another Space
I am using multiple "Mission Control Spaces." When I hit the menu bar item, instead of the window popping up on the active Space, my Mac jumps to another arbitrarily Space.

This is actually a problem that I'm facing with many apps, ever since "upgrading" to Ventura. Is anyone else here experiencing this issue? I'm slowly going insane, and none of my other popup apps have active enough user communities to determine whether this issue is widespread, or just me and some MacOS plist file…

screen cap: https://user-images.githubusercontent.com/617762/216146490-0b4ebbe5-ee32-4f01-9798-95ae6e3d8b5c.mp4

thanks!

tmcknight:
Clicking the tray icon from a fullscreen Space switched to a Desktop Space instead of drawing the popup over the fullscreen app. Ordinary Desktop Spaces were unaffected, which is why this only reproduced against fullscreen apps.

The cause is window level, not Space membership. `electron-menubar` elevates the window unconditionally on Windows, but elsewhere only when the window is kept open on blur:

```js
if (process.platform === "win32") win.setAlwaysOnTop(true, "pop-up-menu");
else win.setAlwaysOnTop(keepOpen || this._options.browserWindow.alwaysOnTop === true);
```

With default settings `keepOpen` is false, so the popup sat at normal window level. The library already passes `visibleOnFullScreen`, so the window may join a fullscreen Space, but a normal-level window still cannot render above one and macOS switches to a Space where it can. Opting into `alwaysOnTop` supplies the missing level; the popup still hides on blur, so it floats only while open.

macOS only. Windows never reads this option, and on Linux an absent key leaves the current `setAlwaysOnTop(false)` behavior unchanged — the new test asserts that, and CI runs on `ubuntu-latest`. Verified manually on macOS 27.0 (arm64) with a packaged build; lint, types, and the 1524 unit tests pass, and no renderer markup changed.

Fixes #563.


tmcknight:
Still need to do some digging.

tmcknight:
Closing as this doesn't fix the issue I'm seeing. I'll post a followup comment on #563 with my findings.

